PR #17, opened 2025-04-28
WIP: What if chained tags didn't need "+k8s:"

thockin:
This is just an idea - looking for feedback. Better or worse?

In this model, the "+k8s:" is an indication that "the rest of this line is a k8s tag" - the tags themselves are named more simply.

### DIFF
--- a/staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/cross_pkg/doc.go
+++ b/staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/cross_pkg/doc.go
@@ -87,22 +87,22 @@ type T1 struct {
 	OtherStructPtr *other.StructType `json:"otherStructPtr"`
 
 	// +k8s:validateTrue="field T1.SliceOfOtherStruct"
-	// +k8s:eachVal=+k8s:validateTrue="field T1.SliceOfOtherStruct values"
-	// +k8s:eachVal=+k8s:opaqueType
+	// +k8s:eachVal=validateTrue="field T1.SliceOfOtherStruct values"
+	// +k8s:eachVal=opaqueType
 	SliceOfOtherStruct []other.StructType `json:"sliceOfOtherStruct"`
 
 	// +k8s:validateTrue="field T1.ListMapOfOtherStruct"
-	// +k8s:eachVal=+k8s:validateTrue="field T1.SliceOfOtherStruct values"
+	// +k8s:eachVal=validateTrue="field T1.SliceOfOtherStruct values"
 	// +k8s:listType=map
 	// +k8s:listMapKey=stringField
-	// +k8s:eachVal=+k8s:opaqueType
+	// +k8s:eachVal=opaqueType
 	ListMapOfOtherStruct []other.StructType `json:"listMapOfOtherStruct"`
 
 	// +k8s:validateTrue="field T1.MapOfOtherStringToOtherStruct"
-	// +k8s:eachKey=+k8s:validateTrue="field T1.MapOfOtherStringToOtherStruct keys"
-	// +k8s:eachVal=+k8s:validateTrue="field T1.MapOfOtherStringToOtherStruct values"
-	// +k8s:eachKey=+k8s:opaqueType
-	// +k8s:eachVal=+k8s:opaqueType
+	// +k8s:eachKey=validateTrue="field T1.MapOfOtherStringToOtherStruct keys"
+	// +k8s:eachVal=validateTrue="field T1.MapOfOtherStringToOtherStruct values"
+	// +k8s:eachKey=opaqueType
+	// +k8s:eachVal=opaqueType
 	MapOfOtherStringToOtherStruct map[other.StringType]other.StructType `json:"mapOfOtherStringToOtherStruct"`
 }
 
@@ -112,11 +112,11 @@ type T1 struct {
 
 /*
 // +k8s:validateTrue="type TypedefSliceOther"
-// +k8s:eachVal=+k8s:opaqueType
+// +k8s:eachVal=opaqueType
 type TypedefSliceOther []other.StructType
 
 // +k8s:validateTrue="type TypedefMapOther"
-// +k8s:eachKey=+k8s:opaqueType
-// +k8s:eachVal=+k8s:opaqueType
+// +k8s:eachKey=opaqueType
+// +k8s:eachVal=opaqueType
 type TypedefMapOther map[other.StringType]other.StructType
 */
--- a/staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/maps/keys/doc.go
+++ b/staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/maps/keys/doc.go
@@ -30,23 +30,23 @@ type Struct struct {
 	TypeMeta int
 
 	// +k8s:validateFalse="field Struct.MapField"
-	// +k8s:eachKey=+k8s:validateFalse="field Struct.MapField(keys)"
+	// +k8s:eachKey=validateFalse="field Struct.MapField(keys)"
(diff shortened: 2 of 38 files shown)


jpbetz:
No concerns with the strategy used here.  Happy to LGTM once rebased and WIP is cleared.

thockin:
Looking at this again, there's some ambiguity now with things like:

```
// +k8s:listType=map
// +k8s+:eachVal=immutable
```

"map" is a leaf value (and a magic word), but not a validator
"immutable" is a validator

What if we ever ended up with a `+k8s:map` tag - it would now be a little more ambiguous.

Do we think that's a problem?

Options:

1) `// +k8s:listType="map"`
2) `// +k8s+:eachVal=+immutable`
3) `// +k8s+:eachVal=+immutable()`
?

jpbetz:
> Options:
> 
> 1. `// +k8s:listType="map"`
> 2. `// +k8s:eachVal=+immutable`
> 3. `// +k8s:eachVal=+immutable()`
>    ?

I'm a big supporter of (1).

I have a strong bias against the use of bare strings.  Every system I've seen use suffered for it.  I realize we've got to grandfather in some tags that have it, but it looks like we can grandfather in quoted strings and still ratchet in the quoted string requirement for all our DV tags?

thockin:
Rebased and updated.  I am still not sure I like this - the inconsistency bothers me a bit.

jpbetz:
> Rebased and updated. I am still not sure I like this - the inconsistency bothers me a bit.

Maybe hold on this till post 1.34?  We're not add enough tags this release to make make it difficult to switch approaches, and we'll have more experience using the tags which may help inform this decision.

thockin:
Rebased.  I am still not sure if this is better or worse.  I am inclined to close it.

jpbetz:
I think we should close it.  We're need to backtrack through https://github.com/kubernetes/gengo/pull/304 if we decided to do this, at this point.

thockin:
I think it works as it is without furhter gengo work. The question is whether the elimination of nested "k8s:" makes DV easier or harder to understand.

jpbetz:
> I think it works as it is without furhter gengo work. The question is whether the elimination of nested "k8s:" makes DV easier or harder to understand.

I find it slightly easier to read. The main benefit to readability I see is that our tags are single-line, and this helps to economizes our use of the limited available space.

The only risk I see is that we can't mix namespaces in a tag.  The whole tag line is "k8s".  I think this is acceptable, but wanted to call it out.

thockin:
Agree that brevity is nice.

OTOH, it means that someone who wants to find all users of a tag needs to hunt for both '+k8s:whatever' and '+k8s:.*=+whatever' which is kind of annoying.

I rebased it just to keep it clean, but I am still slightly inclined to close it, unless we want to rething the syntax even more.  For example, Go's own "directive" syntax is:

```
// This is a regular comment
//go:directive
func Whatever() {
```

should we aim for something more consistent with that?  E.g.

```
// This is a regular comment
//k8s:+optional
//k8s:+subfield(x)=+format=k8s-long-name
```

Unfortunately, most tools special-case `//go:` and not `//.*:`.

e.g.:

```
$ cat test.go
package foo

import "fmt"

// Foo is a function.
//
// k8s:+optional
//
// go:has a space after slashes
//
//go:k8s:+optional
//go:noinline
//go:something I made up
func Foo() {
	fmt.Println("vim-go")
}
```

That was formatted by goimports/gofmt.

```
$ go doc -all .
package foo // import "k8s.io/kubernetes"


FUNCTIONS

func Foo()
    Foo is a function.

    k8s:+optional

    go:has a space after slashes

```

thockin:
ACTUALLY...

https://go-review.googlesource.com/c/go/+/224737

```
$ cat test.go 
package foo

import "fmt"

// Foo is a function.
//
//k8s:optional
//k8s:whatever we want as long as it starts with an alphanumeric
func Foo() {
	fmt.Println("vim-go")
}

$ go doc -all .
package foo // import "k8s.io/kubernetes"


FUNCTIONS

func Foo()
    Foo is a function.

```

Does "godoc automatically ignores it" give us enough incentive to shift syntax?  The `+` makes things more clear, but is it necessary?  We need if in the tag grammar to distinguish naked-string and bool values from chained tags.  Would this be terrible:

```
// This is a regular comment
//
//k8s:optional
//k8s:subfield(x)=+format=k8s-long-name
Spec FooSpec
```

We could change the `+` to `@` or `$` if it was clearer.  We're making this syntax up - before it goes too far, this may be our last chance to be "more standard".

jpbetz:
> I am hoping to avoid matching pairs

Agreed. (our grammar does not demand it)

> //k8s:subfield(x)=@format=k8s-long-name

+1 completely-subjective-personal-preference vote from me (But I don't feel strongly)

thockin:
LOL at myself rediscovering the same bug.  https://github.com/golang/go/issues/59529

thockin:
OK. After investigation, to use directives "properly", we would need to fix Go, and the fix is not obvious.  It's pretty tangled.

We CAN still consider using the standard syntax: 
  1) it might eventually be fixed in Go
  2) it (hiding directives) works in most places (just not the one we need the most)
  3) our custom format (`+directive` or `+k8s:directive`) is not hidden either, so it's not worse than today

Opinions welcome (note: this PR does not implement this idea yet)

thockin:
Curious why support both?  We can convert everything in tree to the new style.  Gengo itself can extract both styles, but we don't need it to be transparent.

E.g. All the new validation-gen tags can be new-style since there's no ecosystem around it.  After that, we can convert each other tool, one by one.  If we are concerned about the broader ecosystem consuming our tags, then I guess we need to keep both styles in the code indefinitely, which mostly defeats the purpose.  This is not something I thought about when I wrote this idea.  It's easy to say "new tags, new syntax".  Much harder to drag the ecosystem along with us.

liggitt:
I'm mostly thinking of out of tree consumers that use types from k8s.io/apimachinery, k8s.io/api, and example.com/some-framework-api, and their own repo. We may want to leave the existing `// +k8s:` tags duplicated / present on *types* for a minor before dropping them to ease upgrade / transition.

I guess it would be fine for new tags to only be added with `//k8s:`, and 1.35 gengo / code-generators to only support `//k8s:` tags (ideally with warnings/errors if non-honored `// +k8s:` tags are detected)

jpbetz:
We have so few +k8s tags in the main Kubernetes repo today (2 in RC, 2 in CSR?) that I think it's fine to just change them (I'd be shocked if anyone has accumulated an actual dependency on these couple tags).

But if we do it, we should do it soon, if we wait till after 1.35, we'd probably need to keep the old tags around.

thockin:
Joe, are you also considering things like:

pkg/apis/core/v1/doc.go:
```
// +k8s:conversion-gen=k8s.io/kubernetes/pkg/apis/core    
// +k8s:conversion-gen-external-types=k8s.io/api/core/v1    
// +k8s:defaulter-gen=TypeMeta    
// +k8s:defaulter-gen-input=k8s.io/api/core/v1    
// +k8s:validation-gen=TypeMeta    
// +k8s:validation-gen-input=k8s.io/api/core/v1  
```

Which could become:

```
//k8s:conversion-gen=k8s.io/kubernetes/pkg/apis/core    
//k8s:conversion-gen-external-types=k8s.io/api/core/v1    
//k8s:defaulter-gen=TypeMeta    
//k8s:defaulter-gen-input=k8s.io/api/core/v1    
//k8s:validation-gen=TypeMeta    
//k8s:validation-gen-input=k8s.io/api/core/v1  
```

Plus the existing tags like `+optional` and `+default=` -- do we retain those forever?

> Long-term, I would be sad if our stuff ended up with half //k8s:... tags for validation and half // +... or // +k8s:... comments for other generators

Agree - half and half would be unfortunate.  I don't know how many other things in the world parse our tags...

PR #22, opened 2025-05-07
[DNM] PoC of using field references in tags to handle minimum and maximum

jpbetz:


jpbetz:
My notes after having tried this out:

- Hoisting a field level validation to the parent struct is the way to go for ratcheting, I don't think passing the parent down like shown here the right way to do it.  Hoisting is a bit messy.  And accessing and setting field paths needs to happen in the validator... Feels like we could ease that via utils and tweaks to the tag handling.
- refs to other values as validation tag values seems tractable.  There are details around how to represent the refs and what limits we place on them is really at the core of this problem.  "$parent" is gross and something I'd like to avoid.   Access to child fields should be limited to basic fields paths "a", "a.b", "a.b.c" and MAYBE listType=map keys ("a[{key1: val1}].b").  It might be best to offer a way to access the struct of a field to provide sibling access in such a way that "$parent" can be avoided.

Issue #29, opened 2025-05-09
[P0] [TODO] Implement Ratcheting Semantics for Update Validation


aaron-prindle:
Design + implement default ratcheting for validation-gen and provide per-validator hooks for advanced ratcheting rules (ex: Ip ratcheting, etc.).

PR #55, opened 2025-05-12
feat: add default ratcheting support for validation-gen

yongruilin:
* equivalence check before emit validators
* ratcheting test cases

ref: https://github.com/jpbetz/validation-gen/issues/29


staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/ratcheting/default_behavior/zz_generated.validations.go
```diff
@@ -0,0 +1,90 @@
...
+	return errs
+}
+
+func Validate_Struct(ctx context.Context, op operation.Operation, fldPath *field.Path, obj, oldObj *Struct) (errs field.ErrorList) {
+	// field Struct.TypeMeta has no validation
+
+	// field Struct.IntField
+	errs = append(errs,
+		func(fldPath *field.Path, obj, oldObj *int) (errs field.ErrorList) {
+			if op.Type == operation.Update && equality.Semantic.DeepEqual(obj, oldObj) {
```
thockin, line 62: You know what we need - a quick benchmark.   

```
package main

import (
	"reflect"
	"testing"

	"sigs.k8s.io/randfill"
)

type T struct {
	I  int
	S  string
	SL [16]*T
}

func BenchmarkEqualInt(b *testing.B) {
	var t1 int
	var t2 int

	fz := randfill.NewWithSeed(1).NilChance(0.5).MaxDepth(8)
	fz.Fill(&t1)
	fz = randfill.NewWithSeed(1).NilChance(0.5).MaxDepth(8)
	fz.Fill(&t2)

	b.ReportAllocs()
	for i := 0; i < b.N; i++ {
		if t1 != t2 {
			b.Fatal("not equal")
		}
	}
}

func BenchmarkDeepEqualInt(b *testing.B) {
	var t1 int
	var t2 int

	fz := randfill.NewWithSeed(1).NilChance(0.5).MaxDepth(8)
	fz.Fill(&t1)
	fz = randfill.NewWithSeed(1).NilChance(0.5).MaxDepth(8)
	fz.Fill(&t2)

	b.ReportAllocs()
	for i := 0; i < b.N; i++ {
		if !reflect.DeepEqual(t1, t2) {
			b.Fatal("not equal")
		}
	}
}

func BenchmarkEqualString(b *testing.B) {
	var t1 string
	var t2 string

	fz := randfill.NewWithSeed(1).NilChance(0.5).MaxDepth(8)
	fz.Fill(&t1)
	fz = randfill.NewWithSeed(1).NilChance(0.5).MaxDepth(8)
	fz.Fill(&t2)

	b.ReportAllocs()
	for i := 0; i < b.N; i++ {
		if t1 != t2 {
			b.Fatal("not equal")
		}
	}
}

func BenchmarkDeepEqualString(b *testing.B) {
	var t1 string
	var t2 string

	fz := randfill.NewWithSeed(1).NilChance(0.5).MaxDepth(8)
	fz.Fill(&t1)
	fz = randfill.NewWithSeed(1).NilChance(0.5).MaxDepth(8)
	fz.Fill(&t2)

	b.ReportAllocs()
	for i := 0; i < b.N; i++ {
		if !reflect.DeepEqual(t1, t2) {
			b.Fatal("not equal")
		}
	}
}

func BenchmarkEqualStruct(b *testing.B) {
	var t1 T
	var t2 T

	fz := randfill.NewWithSeed(1).NilChance(0.5).MaxDepth(8)
	fz.Fill(&t1)
	fz = randfill.NewWithSeed(1).NilChance(0.5).MaxDepth(8)
	fz.Fill(&t2)

	b.ReportAllocs()
	for i := 0; i < b.N; i++ {
		if t1 != t2 {
			b.Fatal("not equal")
		}
	}
}

func BenchmarkDeepEqualStruct(b *testing.B) {
	var t1 T
	var t2 T

	fz := randfill.NewWithSeed(1).NilChance(0.5).MaxDepth(8)
	fz.Fill(&t1)
	fz = randfill.NewWithSeed(1).NilChance(0.5).MaxDepth(8)
	fz.Fill(&t2)

	b.ReportAllocs()
	for i := 0; i < b.N; i++ {
		if !reflect.DeepEqual(t1, t2) {
			b.Fatal("not equal")
		}
	}
}
```

```
BenchmarkEqualInt-6          	1000000000	         0.4059 ns/op	       0 B/op	       0 allocs/op
BenchmarkDeepEqualInt-6      	25531468	        46.41 ns/op	      16 B/op	       2 allocs/op
BenchmarkEqualString-6       	294807688	         3.737 ns/op	       0 B/op	       0 allocs/op
BenchmarkDeepEqualString-6   	16466414	        74.84 ns/op	      32 B/op	       2 allocs/op
BenchmarkEqualStruct-6       	99291078	        10.46 ns/op	       0 B/op	       0 allocs/op
BenchmarkDeepEqualStruct-6   	 2330167	       523.0 ns/op	     320 B/op	       2 allocs/op
```

DeepEqual is 25-100x slower.  More importantly it saves a couple of allocations per-field.  That could matter a lot.

Granted, this is the microest of micro-benchmarks, but it says to me that we should TRY to emit direc comparison code when we can.  Even if that is just for Builtins (we can always add the richer idea of deep-comparable later).

staging/src/k8s.io/code-generator/cmd/validation-gen/validation.go
```diff
@@ -1085,6 +1087,13 @@ func (g *genValidations) emitCallToOtherTypeFunc(c *generator.Context, node *typ
 // variables named "obj" and "oldObj", and the field path to this value is
 // named "fldPath".
 func emitCallsToValidators(c *generator.Context, validations []validators.FunctionGen, sw *generator.SnippetWriter) {
+	targs := generator.Args{
+		"operation": mkSymbolArgs(c, operationPkgSymbols),
+		"equality":  mkSymbolArgs(c, equalityPkgSymbols),
+	}
```
thockin, line 1093: if you pass the type in, then you can check it for builtin (remember to unalias and unpointer it) and instead just do "==".

Edit: I guess you need to handle pointers and check for nil, too.  A little more complicated but not a ton.

staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/ratcheting/default_behavior/doc.go
```diff
@@ -0,0 +1,41 @@
...
+// +k8s:validation-gen-scheme-registry=k8s.io/code-generator/cmd/validation-gen/testscheme.Scheme
+
+// This is a test package.
+package defaultbehavior
+
+import "k8s.io/code-generator/cmd/validation-gen/testscheme"
+
+var localSchemeBuilder = testscheme.New()
+
+type Struct struct {
```
thockin, line 27: I think you need these cases plus equivalent cases which are not direct-comparable (slices work well for this).

Also, each attachment-point should have 2 tags (can both be validateFalse) so we prove that the `if` is only emitted once.

thockin: Also, validateFalse lets you put a useful string in the value, so your test can ensure that it got the RIGHT failure.

staging/src/k8s.io/code-generator/cmd/validation-gen/validation.go
```diff
@@ -1085,6 +1087,13 @@ func (g *genValidations) emitCallToOtherTypeFunc(c *generator.Context, node *typ
...
+	sw.Do("if op.Type == $.operation.Update|raw$ && $.equality.Semantic|raw$.DeepEqual(obj, oldObj) {\n", targs)
```
aaron-prindle, line 1094: It might make sense to add a comment explaining that this generated ratcheting code is only for the general case.  Additionally IIUC this PR is for the general case but how does this code work for cases where obj is a list of listType=atomic and re-ordering uses a custom approach vs DeepEquals?  I believe this is called out in the TODOs and planned to be addressed in a separate PR but was wondering if this code currently handles this?  Might make sense to add a comment around this as well.

staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/ratcheting/default_behavior/doc.go
yongruilin: I've added more test cases and change the tag to validateFalse

staging/src/k8s.io/code-generator/cmd/validation-gen/validation.go
yongruilin: Added a new emit func `emitRatchetingCheck(ctx, type, sw)`.
Since we added the special handling for the map/slice in https://github.com/jpbetz/validation-gen/pull/24, we should not check equivalence  every time when `emitCallsToValidators`.

staging/src/k8s.io/code-generator/cmd/validation-gen/validation.go
yongruilin: Added a new func to wrap the emitting of ratcheting with some comments.

staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/ratcheting/default_behavior/zz_generated.validations.go
jpbetz: I came to the same conclusion: https://github.com/jpbetz/validation-gen/pull/55#pullrequestreview-2844005102

It looks like it simplifies the code as well.  Double win.

staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/ratcheting/default_behavior/doc.go
```diff
@@ -0,0 +1,66 @@
...
+
+	// +k8s:validateFalse="field intField"
+	IntField int `json:"intField"`
+
+	// +k8s:optional
+	// +k8s:validateFalse="field intPtrField"
+	IntPtrField *int `json:"intPtrField"`
+
+	// +k8s:validateFalse="field typedefStruct"
+	TypeDefStruct AliasStruct `json:"typedefStruct"`
```
jpbetz, line 38: nit:  Let's not mix the terms typeref and alias since go type definitions and type alias are different concepts.
```suggestion
	AliasStruct AliasStruct `json:"typedefStruct"`
```

jpbetz:
First off- I love how simple this is.

Let's continue to expand test coverage further.

I see we're covering:

- A selection of tags (forbidden, format, maxItems, maxLength, minimum, required, validateFalse>
- A selection of primitive combinations:
  - `<scalar>`
  - `*<scalar>`
  - `$typedef<struct>`
  - `$typedef<[]<scalar>>`
  - `map[string]$typedef<string>`
  - `[]<primitive>`
  - `[]<struct>`
  - `$typedef<[]<scalar>>`

I'd like our testing rigor to be increased in two main ways:

- Extensive coverage of of our main "atoms" (comparable scalar, comparable pointer-to-scalar, a non-comparable value (Quantity, metav1.Time or similar) and non-comparable pointer-to-value) composed with: struct, map, slice, typedef, type alias (we do have type aliases in the API) ...   I'd like there to specifically be cases that are still comparable (struct containing all possible comparable types, including a nested struct and a pointer to a comparable).   Should be >>16 test cases to cover that.

- A plan for how we ensure that future tags are tested for ratcheting.  I'd expect the tests for `union`, `subfield` and `opaque` to set the bar for this.  If this PR doesn't cover this because it's beyond default ratcheting that's OK, I'd just like to make sure we track and fully handle it.

staging/src/k8s.io/code-generator/cmd/validation-gen/validation.go
```diff
@@ -1071,6 +1091,27 @@ func (g *genValidations) emitCallToOtherTypeFunc(c *generator.Context, node *typ
...
+	// TODO: Need to have a follow-up PR to handle the default behavior
+	// for slices and maps.
+	targs := generator.Args{
+		"operation": mkSymbolArgs(c, operationPkgSymbols),
+	}
+	// If the type is a builtin, we can use a simpler equality check when they are not nil.
+	if validators.NonPointer(validators.NativeType(t)).Kind == types.Builtin {
+		_, exprPfx, _ := getLeafTypeAndPrefixes(t)
+		targs["exprPfx"] = exprPfx
+		sw.Do("if op.Type == $.operation.Update|raw$ && (obj == oldObj || (obj != nil && oldObj != nil && $.exprPfx$obj == $.exprPfx$oldObj)) {\n", targs)
```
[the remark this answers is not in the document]
yongruilin: I insert a commit to implement a isDirectComparable(t types.Type) for it.

staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/ratcheting/default_behavior/doc.go
```diff
@@ -0,0 +1,109 @@
...
+	MapValueField map[string]S `json:"mapValueField"`
+
+	// +k8s:validateFalse="field aliasMapKeyTypeField"
+	AliasMapKeyTypeField AliasMapKeyType `json:"aliasMapKeyTypeField"`
+
+	// +k8s:validateFalse="field aliasMapValueTypeField"
+	AliasMapValueTypeField AliasMapValueType `json:"aliasMapValueTypeField"`
+}
+
+type StructStruct struct {
```
jpbetz, line 64: Can we also test that an embedded struct (both direct comparable and non-direct comparable) result in the correct comparison type for the containing struct?

yongruilin: Added the test cases for embedded struct.

jpbetz: Is this sufficient?  I was expecting a dedicated struct containing an embedded DirectComparableStruct that could be used to confirm that "a struct embedded a direct comparable struct is also directly comparable"

jpbetz:
LGTM

Leaving merge to @aaron-prindle

staging/src/k8s.io/code-generator/cmd/validation-gen/validation.go
```diff
@@ -1459,3 +1459,24 @@ func (g *fixtureTestGen) Init(c *generator.Context, w io.Writer) error {
...
+	case types.Builtin:
+		return true
+	case types.Struct:
+		for _, f := range t.Members {
+			if !isDirectComparable(f.Type) {
+				return false
+			}
+		}
+		return true
+	case types.Alias:
```
aaron-prindle, line 1519: I believe that types.Array are also directly comparable for completeness (when compared with similarly typed array) for completeness.  I don't believe fixed length arrays are used commonly (at all?) in K8s APIs though as slices are preferred so it is likely fine to keep the DeepEquals for this case.

staging/src/k8s.io/code-generator/cmd/validation-gen/validation.go
```diff
@@ -1071,6 +1091,27 @@ func (g *genValidations) emitCallToOtherTypeFunc(c *generator.Context, node *typ
 	sw.Do("errs = append(errs, $.funcName|raw$(ctx, op, fldPath, obj, oldObj)...)\n", targs)
 }
 
+// emitRachetingCheck emits a equivalence check for default ratcheting.
+func emitRatchetingCheck(c *generator.Context, t *types.Type, sw *generator.SnippetWriter) {
+	// Emit equavilance check for default ratcheting.
```
aaron-prindle, line 1096: nit: s/equavilance/equivalence

staging/src/k8s.io/code-generator/cmd/validation-gen/validation.go
```diff
@@ -1071,6 +1091,27 @@ func (g *genValidations) emitCallToOtherTypeFunc(c *generator.Context, node *typ
...
+	// TODO: Need to have a follow-up PR to handle the default behavior
```
aaron-prindle, line 1097: Can you create an issue around this work item and have the TODO reference the issue:
```go
// TODO(#123): ...
```

This way we can track remaining work items better and potentially have work items broader community folks could work on or other teammembers over time

staging/src/k8s.io/code-generator/cmd/validation-gen/validation.go
yongruilin: Updated thanks!

staging/src/k8s.io/code-generator/cmd/validation-gen/validation.go
yongruilin: We don't use Array in k8s api. But I updated the function to handle the Array as comparable.

staging/src/k8s.io/code-generator/cmd/validation-gen/validation.go
yongruilin: Thanks, I think I will omit the issue number here since we are still baking it in the dev repo. We can add the issue in k/k if the TODO is still needed when to send the PR to k/k.

PR #56, opened 2025-05-13
[DNM] [WIP] feat: add list element access to +k8s:subfield

aaron-prindle:
The PR extends +k8s:subfield(subname) to also support list map element selection  with a new syntax (see example below):

```
type Struct struct {
	// +k8s:subfield({"listElems":{"type":"Approved"}})=+k8s:validateFalse="subfield Conditions[type=Approved]"
	// +k8s:subfield({"listElems":{"type":"NonExistent"}})=+k8s:validateFalse="subfield Conditions[type=NonExistent]"
	// +k8s:subfield({"listElems":{"status":"True", "type":"Approved"}})=+k8s:validateFalse="subfield Conditions[status=True,type=Approved]"
	// +k8s:subfield({"listElems":{"stringPtr":"Target"}})=+k8s:validateFalse="subfield Conditions[stringPtr=Target]"
	Conditions []MyCondition `json:"conditions"`
}

type MyCondition struct {
	Type      string  `json:"type"`
	Status    string  `json:"status"`
	StringPtr *string `json:"stringPtr"`
}

```

The goal of this PR is to lay the groundwork for supporting +k8s:unionMember on the "virtual" list map element field that subfield select which is planned to be part of the CSR migration for v1.34 :
```go
       // +listType=map
       // +listMapKey=type
       // +optional
       // +ks8:subfield({"type":"Approved", "status": "true"})=+k8s:unionMember
       // +ks8:subfield({"type":"Denied", "status": "true"})=+k8s:unionMember
	Conditions []CertificateSigningRequestCondition
```

Notes related to the PR:
- This PR requires a gengo change to allow for JSON related character (`{`, `}`, `,`,`:`, `"`etc.) to be put into the gengo "args".  This PR has a naive version of the necessary gengo change included but such a change would need to  be upstreamed in gengo prior to merging this and removed from this PR
- For experimentation this PR currently has the syntax +k8s:subfield({"listElems": {...},"Flags":{...}}) to attempt to experiment with No error and ShortCircuit.  This will likely be removed so the synax has listElems as the top level JSON payload (eg: +k8s:subfield({keyValues...}))
- While it is possible to chain +k8s:subfield({...})=+k8s:subfield(subname)=+k8s:required, it is not possible directly to chain +k8s:subfield({...})=+k8s:required.  This will result in a panic as Member will be nil. Such combinations should be prevented at generation time.  This inabiliity makes sense though as +k8s:required for such a virtual field doesn't make sense directly (+k8s:unionMember does make sense).


thockin:
A few questions/comments before I even read it.

1) This is kind of awesome.
2) Am I wrong to interpret this as a sort of "select-where" clause?  `select * from field where ...` 
3) IIUC this is an exact-match, with no compound or negative matches?  E.g. no `select * where type != "foo"` ?  How likely is that to survive before we hit a need for something more?
4) It feels unbalanced to not have an equivalent for maps.  What would the syntax be?
5) "it is not possible directly to chain +k8s:subfield({...})=+k8s:required. This will result in a panic as Member will be nil" - that seems solvable to me
6) Does it only work for scalar values or do we need object matches and/or list matches?

To points 3 and 4 - I know we decided to back-burner CEL in favor of specific semantic tags, but this feels like another place where CEL is the OBVIOUS choice.  Concretely, I would not want to see us try to model compound booleans or not expressions in our own way.  So I see two possible paths:

1) Do a simple form of exact-match, like you have, and then later add CEL.  I'd want to see us think thru the evolution of it, so we know we can add it.
2) Consider CEL for this now, e.g. 

```
// +k8s:select(`item.type="Approved" && item.status=="True"`)=+unionMember
```

Note: I used `select` as the tag name because it feels like a DB select (with N matches) than a subfield (which implies 1 match) to me, but I am flexible.

Advantages of CEL here:
* don't need a new grammar
* lists and maps both work  (e.g. the labels map might use `item.startsWith("k8s.io/")`)
* sub-structs work:  `item.foo.bar == "Ready"`

staging/src/k8s.io/apimachinery/pkg/api/validate/subfield.go
```diff
@@ -39,3 +45,103 @@ func Subfield[Tstruct any, Tfield any](ctx context.Context, op operation.Operati
...
 }
+
+func ListMapElementByKey[TList ~[]TItem, TItem any](
+	ctx context.Context, op operation.Operation, fldPath *field.Path,
+	newList, oldList TList, getMap GetMapFunc,
+	elementValidator func(ctx context.Context, op operation.Operation, fldPath *field.Path, newObj, oldObj *TItem) field.ErrorList,
+) field.ErrorList {
+
+	var errs field.ErrorList
+	m := getMap()
```
thockin, line 56: This is going to be called for every relevant API operation for each such field.  Maps are deceptively expensive.  Why not pass a match function, like:

```
type MatchFn[T any] func(T) bool
```

Codegen emits a FunctionLiteral which is like `func(x Condition) bool { x.type == "Foo" && x.Value == "True"}`. 

Then the loop here says `if !match(newList[i]) { continue }`

And if it uses CEL, you just pass the CEL evaluator as the MatchFn ?

staging/src/k8s.io/apimachinery/pkg/api/validate/subfield.go
```diff
@@ -39,3 +45,103 @@ func Subfield[Tstruct any, Tfield any](ctx context.Context, op operation.Operati
...
+			errs = append(errs, elementValidator(ctx, op, elementPathForSelector, nil, &oldList[i])...)
+		}
+	}
+
+	return errs
+}
+
+func matchesAllSelectors[TItem any](item TItem, m map[string]string) bool {
+	for matchK, matchV := range m {
+		if fieldV, ok := getReflectedJSONFieldValueAsString(reflect.ValueOf(item), matchK); !ok || fieldV != matchV {
```
thockin, line 94: We should not need to fall onto reflect, I think.

staging/src/k8s.io/apimachinery/pkg/api/validate/subfield.go
```diff
@@ -39,3 +45,103 @@ func Subfield[Tstruct any, Tfield any](ctx context.Context, op operation.Operati
...
+		// TODO(aaron-prindle) stripping aliasing more similar to current patterns than this way.
+		if fieldValue.Type().ConvertibleTo(reflect.TypeOf("")) {
+			return fieldValue.Convert(reflect.TypeOf("")).String(), true
+		}
+		return "", false
+	}
+	return "", false
+}
+
+func getJSONFieldName(field reflect.StructField) (string, bool) {
```
thockin, line 130: We have some utils for this - we should factor out a single lib that does it "well enough" for everyone :)

staging/src/k8s.io/apimachinery/pkg/api/validate/subfield.go
aaron-prindle: I've updated the code to so that the matching uses the method recommended above passing a:
```go
type MatchFn[TItem any] func(item *TItem) bool
```

and having codegen emit code for the match function like that below (snippet from PR post update)

```
			errs = append(errs, validate.ListMapElementByKey(ctx, op, fldPath, obj, oldObj, func() map[string]string { return map[string]string{"status": "True", "type": "Approved"} }, func(item *MyCondition) bool {
				if item == nil {
					return false
				}
				return item.Status == "True" && item.Type == "Approved" // <--- HERE
			}, func(ctx context.Context, op operation.Operation, fldPath *field.Path, obj, oldObj *MyCondition) field.ErrorList {
				return validate.FixedResult(ctx, op, fldPath, obj, oldObj, false, "subfield Conditions[status=True,type=Approved]")
			})...)
```

Currently though I still do pass a map in a separate function which had the goal of making the field path have information that a selector was used:

```go
// GetSelectorMapFn returns a map - needed because the selector map is
// used for generating the field path.
type GetSelectorMapFn func() map[string]string
```

The idea was that the fldPath could possibly look like the below when this selector is used:
```
conditions[status="True",type="Approved"]
```

staging/src/k8s.io/apimachinery/pkg/api/validate/subfield.go
aaron-prindle: You are correct, I've updated the code to now be more similar to the normal usage of +k8s:subfield where instead of using reflection at runtime, instead codegen outputs the match function which has fields.  Example:


```go
			errs = append(errs, validate.ListMapElementByKey(ctx, op, fldPath, obj, oldObj, func() map[string]string { return map[string]string{"status": "True", "type": "Approved"} }, func(item *MyCondition) bool {
				if item == nil {
					return false
				}
				return item.Status == "True" && item.Type == "Approved" // <--- HERE
			}, func(ctx context.Context, op operation.Operation, fldPath *field.Path, obj, oldObj *MyCondition) field.ErrorList {
				return validate.FixedResult(ctx, op, fldPath, obj, oldObj, false, "subfield Conditions[status=True,type=Approved]")
			})...)
```

staging/src/k8s.io/apimachinery/pkg/api/validate/subfield.go
aaron-prindle: +1, to having some shared utils for JSON -> field and field -> JSON.  This method is actually removed/not-needed now after changing to code from the suggestion in https://github.com/jpbetz/validation-gen/pull/56#discussion_r2087585993.

jpbetz:
> Am I wrong to interpret this as a sort of "select-where" clause? select * from field where ...

This has incredibly broad applications.

For example, if we had `select`, the rule that a CSR may not have both the "Approved" and "Denied" status condition could be expressed using our existing tags as: 

`+k8s:select(`item.type=="Approved" || item.type=="Denied"`)=+k8s:maximum=1`

An alternative to consider would be to support "selectors" as arguments to function type tags that operate on aggregate types.  For example:

`+k8s:maximum(select:`item.type in ["Approved", "Denied"]`)=1`

This suggests that we might not need to handle `subfield(key) + union` to satisfy our CSR use cases.

Note that I do still consider `subfield(<key>)` useful, but it might just be a lightweight alias over this more general mechanism?

jpbetz:
cc @deads2k for https://github.com/jpbetz/validation-gen/pull/56#issuecomment-2877841962 and https://github.com/jpbetz/validation-gen/pull/56#issuecomment-2884293370

jpbetz:
One last thought.  For the `subfield` tag we're introducing here.  Should we scope it down to just access of a properly keyed listType=map?

For status conditions, which we know have `listType=map elements where the validation is aware of particular keys, being able to do `+k8s:subfield(<key>)=<tag we know applies to a single element>` seems particularly useful.  

I bring this up, because if I use a key to access as listType=map,  I think I want validation-gen to error our during generation if I fail to fully specify the key, or if I try to use a non-key field.

Also, what should we name the tag?

Should we coomoditize tags and support field access, listType=map access and map access all though a `+k8s:subfield` tag, or should we have `+k8s:subfield(<field-name>)`, `+k8s:keyedElement(<key-value-pairs>)` and `+k8s:mapValue(<string-key>)` ?

aaron-prindle:
> One last thought. For the subfield tag we're introducing here. Should we scope it down to just access of a properly keyed listType=map?

+1, to this approach. I will modify this PR to extend subfield to allow access to a single properly keyed listType=map.  This would only be able to select up to 1 element, in line with subfield today.

In line with changing the scope of this of PR, I have created two issues against `validation-gen` to continue discussions around:
- Discuss DB style select of N elements (+k8s:select + alternatives): https://github.com/jpbetz/validation-gen/issues/59
- Discuss potential options and preferred method for handling the "union across conditions - [Approved,Denied]" validation rule identified in CSR for v1.34: <issue-incoming-will-edit-this-post>

thockin:
This got long, and I realized Joe said a lot of the same things.

I think it comes down to how we want to compose tags.  I was thinking of `select` and meaning `eachVal` for the matching items.  Joe's framing is more like a list comprehension in Python.  Maybe twist it a bit and get:

```
// This says "each value in the list result is a member of the union".
//+k8s:select('item.type in ["Approved", "Denied"]')=+k8s:eachVal=+k8s:unionMember

// This says "the list result has maxItems".
//+k8s:select('item.type in ["Approved", "Denied"]')=+k8s:maxItems=1
```

Two ways of saying the same thing?

Calling it `select` vs `eachVal` is not just aesthetic.  `eachVal` works for lists and maps already, and `eachKey` would allow you to operate only on keys that match. I'm not sure that's the right meaning though -- I think we want to operate on each value that matches the key, so maybe `select` is better.

My fear is that our zest for generality leads us down a path towards having to create a runtime environment with intermediate representations.  Maybe we should look at what generated (pseudo)code would be?

```
+k8s:maxItems('item.type in ["Approved", "Denied"]')=1
```

would become something like:

```
validate.MaxItems(list, matchFn, max) {
    if matchFn == nil && len(list) > max {
        return error
    }

    found = 0
    for each item in list {
        if matchFn(item) {
            found++
            if found > max {
                return error
            }
        }
    }
}
```

Not too bad.

```
+k8s:eachVal('item.type in ["Approved", "Denied"]')=+k8s:unionMember
```

...would be more complicated to set up but the runtime code is even simpler. Something like:

```
validate.Union(membershipInfo, values ...any) {
    for each v in values {
        if isSpecified(v) {
            ...
        }
    }
}
```

...which is basically what exists today.  The tricky part is changing union logic to be more open-ended wrt number of values presented vs. number of fields in the membership info.

> For the subfield tag we're introducing here. Should we scope it down to just access of a properly keyed listType=map?

I think we have a few possible semantics, we should decide how many of them we need?

1) `subField(<selector>)` implies ZERO-OR-ONE results, operating on the value that matches the selector.  That could work with a full listmap key selector or a map key literal.  I think we would want to ensure it was a full exact-match for the key -- no prefixes or "in list" or partial matching.  An object-literal seems like the right syntax, but we might need to constrain it to just primitives.

2) `eachVal(<selector>)` and `eachKey(<selector>)` implies ZERO-OR-N results, operating on the values or keys (respectively) that match the selector. That could allow partial matches or non-listmap-key fields.  CEL seems like the right syntax.

3) `select(<selector>)` implies ZERO-OR-N results, operating on the values that match the selector. That could allow partial matches or non-listmap-key fields. CEL seems like the right syntax.

4) `maxItems(<selector>)` seems like a different way of doing #2 or #3.

staging/src/k8s.io/apimachinery/pkg/api/validate/subfield.go
```diff
@@ -39,3 +42,44 @@ func Subfield[Tstruct any, Tfield any](ctx context.Context, op operation.Operati
...
+				oldJ = &oldList[j]
+				processedOldIndices[j] = true
+				break
+			}
+		}
+		// Pass matching element present in newList and oldList to validator.
+		errs = append(errs, elementValidator(ctx, op, fldPath.Index(i), curI, oldJ)...)
+	}
+
+	for i := range oldList {
```
thockin, line 77: I'm unclear what this is doing - we don't call any other validations on removed items -- do we need to?

staging/src/k8s.io/apimachinery/pkg/api/validate/util/util.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"fmt"
+	"sort"
+	"strings"
+)
+
+// GeneratePathForMap generates sorted k,v entries as a string from a map - 'status="true",type="Approved"'
+// This is used as part of the +k8s:subfield context path for some cases, ex:
+// Struct.Conditions[status="true",type="Approved"]
+func GeneratePathForMap(keyValues map[string]string) string {
```
thockin, line 28: We should avoid maps if we can.  `[]string` or `[[2]string` ?

staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/tags/subfield/listsimple/doc.go
```diff
@@ -0,0 +1,40 @@
...
+package listsimple
+
+import "k8s.io/code-generator/cmd/validation-gen/testscheme"
+
+var localSchemeBuilder = testscheme.New()
+
+type Struct struct {
+	TypeMeta int `json:"typeMeta"`
+
+	// +k8s:subfield({"listElems":{"type":"Approved"}})=+k8s:validateFalse="subfield Conditions[type=Approved]"
```
thockin, line 29: If this list isn't a listmap or doesn't have keys, should such a subfield fail?

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/subfield.go
```diff
@@ -49,67 +54,266 @@ func (subfieldTagValidator) ValidScopes() sets.Set[Scope] {
...
+				getFn.Body = fmt.Sprintf("return %so.%s", fieldExprPrefix, submemb.Name)
+				f := Function(subfieldTagName, vfn.Flags, validateSubfield, subname, getFn, WrapperFunction{vfn, submemb.Type})
+				result.Functions = append(result.Functions, f)
+				result.Variables = append(result.Variables, validations.Variables...)
 			}
+		}
+		return result, nil
+
+	} else { // +k8s:subfield({"listMapElems":{"..."}}) usage
+		if t.Kind != types.Slice && t.Kind != types.Array {
```
thockin, line 145: simpler errors: `fmt.Errorf("can only be used on list types")`  or maybe "...list-map types" ?

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/subfield.go
```diff
@@ -49,67 +54,266 @@ func (subfieldTagValidator) ValidScopes() sets.Set[Scope] {
...
+			return Validations{}, fmt.Errorf("list access via 'listElems' can only be used on slice/array types, but tag is on field %s of type %s", context.Path.String(), t.Name.String())
+		}
+
+		elemT := nonPointer(nativeType(t.Elem))
+		if elemT.Kind != types.Struct {
+			return Validations{}, fmt.Errorf("elements of slice/array must be structs for key-based selection, but elements of field %s are %s", context.Path.String(), elemT.Name.String())
```
thockin, line 151: same - simpler

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/subfield.go
```diff
@@ -49,67 +54,266 @@ func (subfieldTagValidator) ValidScopes() sets.Set[Scope] {
...
+		// generates context path like Struct.Conditions[status="true",type="Approved"]
+		subContextPath := util.GeneratePathForMap(parsedArg.ListElems)
+		subContext := Context{
+			Scope: ScopeField,
+			Type:  elemT,
+			// TODO(aaron-prindle) for +k8s:unionMember support need to plumb this.
+			Parent: nil,
+			Path:   context.Path.Key(subContextPath),
+			// TODO(aaron-prindle) for +k8s:unionMember support need to plumb this.
+			Member: nil,
```
thockin, line 173: Putting Member into context was "easy" but it may be that we want our own struct which has exaclty the info we need and can be used for both fields and list-map members.

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/subfield.go
```diff
@@ -49,67 +54,266 @@ func (subfieldTagValidator) ValidScopes() sets.Set[Scope] {
...
-				Parameters: []ParamResult{{"o", nilableStructType}},
-				Results:    []ParamResult{{"", nilableFieldType}},
+		for k := range parsedArg.ListElems {
+			if getMemberByJSON(elemT, k) == nil {
+				return Validations{}, fmt.Errorf("element type %s (of list %s) has no field with JSON name %q", elemT.Name.String(), context.Path.String(), k)
 			}
-			getFn.Body = fmt.Sprintf("return %so.%s", fieldExprPrefix, submemb.Name)
-			f := Function(subfieldTagName, vfn.Flags, validateSubfield, subname, getFn, WrapperFunction{vfn, submemb.Type})
-			result.Functions = append(result.Functions, f)
+		}
```
thockin, line 158: need to ensure the field is a builtin or string?

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/subfield.go
```diff
@@ -49,67 +54,266 @@ func (subfieldTagValidator) ValidScopes() sets.Set[Scope] {
...
+
```
thockin, line 159: need to ensure they specified the whole listmap key?  To get that info, you might need to make this closer to `each.go` tags than subfield.  It's cute to use the same tag name, but not really NECESSARY, right?

thockin (commented):
Ran out of time to dig deeper on it for now.  Lots to think about.

staging/src/k8s.io/apimachinery/pkg/api/validate/subfield.go
```diff
@@ -39,3 +42,44 @@ func Subfield[Tstruct any, Tfield any](ctx context.Context, op operation.Operati
...
+
+	for i := range newList {
+		curI := &newList[i]
+		if !matches(curI) {
+			continue
+		}
+		var oldJ *TItem
+
+		for j := range oldList {
+			if processedOldIndices[j] {
```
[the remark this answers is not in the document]
aaron-prindle: Thanks for flagging this, I have updated this code now and it no longer users processedOldIndices for the reasons mentioned.  This was leftover from a version of this code when listMapKey was not guaranteed to be in the selector and multiple entries were allowed.

staging/src/k8s.io/apimachinery/pkg/api/validate/subfield.go
aaron-prindle: If used with +k8s:immutable - +k8s:subfield({“type”: “Accepted”})=+k8s:immutable my understanding would be that if someone attempts to remove the "immutabel" list entry matching “type”:”Accepted” then this should fail as that entry is immutable once created.

staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/tags/subfield/listsimple/doc.go
aaron-prindle: Yes it should, I have updated the code now to check that the list has // listType=map and that // listMapKey=... matches one of the selector keys

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/subfield.go
aaron-prindle: Done, updated to "can only be used on list types".

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/subfield.go
aaron-prindle: Done

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/subfield.go
aaron-prindle: Done, there is a check to ensure this is a string type now

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/subfield.go
aaron-prindle: > Need to ensure they specified the whole listmap key? To get that info, you might need to make this closer to each.go tags than subfield.  

IIUC this is referring to checking that the listMapKey=foo value is one of the keys in the "matcher map" {"foo": "value"}.  I recently added a check that does this with:
```
listMapKey := parseListMapKey(context.Member.CommentLines)
```

Happy to change/fix this if this method has an issue or this referring to a different change.

staging/src/k8s.io/apimachinery/pkg/api/validate/util/util.go
aaron-prindle: Currently the "matcher map" JSON is parsed into a map:
```
	// +k8s:subfield({"type":"Approved"})=+k8s:validateFalse="subfield Conditions[type=Approved]"
```

```
	var matcherMap map[string]string
	if err := json.Unmarshal([]byte(argStr), &matcherMap); err == nil {
```

```
		subContextPath := util.GeneratePathForMap(parsedArg.MatcherMap)

```

Does it make sense to convert this to []string // [][2]string given this? Alternatively would it make sense to change the input format to avoid map usage?

aaron-prindle:
From prior offline discussion I have updated the PR here to no longer extend +k8s:subfield but instead implement a new tag for this functionality - +k8s:subListMapItem.  Feel free to add any naming suggestions as well.  From the PR currently I believe things are in a good with the only open question being the best way to pass the arguments to this tag.  Currently this uses a JSON map in the gengo args:
```go
	// +listType=map
	// +listMapKey=type
	// +k8s:subListMapItem({"status":"True","type":"Approved"})=+k8s:validateFalse="subListMapItem Conditions[status=True,type=Approved]"
	Conditions []MyCondition `json:"conditions"`
```

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/sublistmapitem.go
```diff
@@ -0,0 +1,290 @@
...
+	return "", fmt.Errorf("no field with JSON name %q in type %s", jsonKey, s.Name.String())
+}
+
+func parseListMapKey(commentLines []string) string {
+	var key string
+	hasListTypeMap := false
+
+	for _, line := range commentLines {
+		trimmedLine := strings.TrimSpace(line)
+		if trimmedLine == "+listType=map" {
```
yongruilin, line 237: Should we use "+k8s:listType=map" here to match the validators/each.go ?

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/sublistmapitem.go
```diff
@@ -0,0 +1,290 @@
...
+
+	if context.Member == nil {
+		// TODO(aaron-prindle) more descriptive error would be helpful here
+		// currently not sure what cases this occurs for
+		return Validations{}, fmt.Errorf("unexpected nil context member")
+	}
+	listMapKey := parseListMapKey(context.Member.CommentLines)
+	if len(listMapKey) == 0 {
+		return Validations{}, fmt.Errorf("must have '+listType=map' and '+listMapKey=...' annotations to use subListMapItem with matcher map")
+	}
```
yongruilin, line 97: This block is a bit similar to the logic in `each.go - listTypeTagValidator` (except each.go is looking for +k8s:xx), I couldn't come up with a good way though, did you consider to couple this new tag with the listMetadata?

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/sublistmapitem.go
```diff
@@ -0,0 +1,290 @@
...
+	if _, ok := parsedArg.MatcherMap[listMapKey]; !ok {
+		return Validations{}, fmt.Errorf("subListMapItem matcher map must contain listMap key")
+	}
```
yongruilin, line 100: listMapKey could have multiple keys. Should all the keys be part of the args?

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/sublistmapitem.go
```diff
@@ -0,0 +1,290 @@
...
+			keyPart := strings.TrimPrefix(trimmedLine, "+listMapKey=")
+			trimmedKey := strings.TrimSpace(keyPart)
+			if trimmedKey != "" {
+				key = trimmedKey
+			}
+		}
+	}
+
+	if hasListTypeMap && key != "" {
+		return key
```
yongruilin, line 250: Maybe multiple keys?

jpbetz:
> I think we have a few possible semantics, we should decide how many of them we need?
> 
> 1. `subField(<selector>)` implies ZERO-OR-ONE results, operating on the value that matches the selector.  That could work with a full listmap key selector or a map key literal.  I think we would want to ensure it was a full exact-match for the key -- no prefixes or "in list" or partial matching.  An object-literal seems like the right syntax, but we might need to constrain it to just primitives.
> 2. `eachVal(<selector>)` and `eachKey(<selector>)` implies ZERO-OR-N results, operating on the values or keys (respectively) that match the selector. That could allow partial matches or non-listmap-key fields.  CEL seems like the right syntax.
> 3. `select(<selector>)` implies ZERO-OR-N results, operating on the values that match the selector. That could allow partial matches or non-listmap-key fields. CEL seems like the right syntax.
> 4. `maxItems(<selector>)` seems like a different way of doing (2) or (3).

+100.  I think this is what we need to decide.  I worry that organically we might end up with all of these unless we guide the effort.

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/sublistmapitem.go
```diff
@@ -0,0 +1,290 @@
...
+
+	elemT := NonPointer(NativeType(t.Elem))
+	if elemT.Kind != types.Struct {
+		return Validations{}, fmt.Errorf("can only be used on list of structs")
+	}
+
+	if context.Member == nil {
+		// TODO(aaron-prindle) more descriptive error would be helpful here
+		// currently not sure what cases this occurs for
+		return Validations{}, fmt.Errorf("unexpected nil context member")
```
jpbetz, line 92: I think you need to change the scope of this tag from any to members given this guard.

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/sublistmapitem.go
jpbetz: +1, a repeated tag may be used to declare multiple key fields

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/sublistmapitem.go
```diff
@@ -0,0 +1,290 @@
...
+func parseListMapKey(commentLines []string) string {
+	var key string
+	hasListTypeMap := false
+
+	for _, line := range commentLines {
+		trimmedLine := strings.TrimSpace(line)
+		if trimmedLine == "+k8s:listType=map" {
+			hasListTypeMap = true
+		}
+		if strings.HasPrefix(trimmedLine, "+k8s:listMapKey=") {
```
jpbetz, line 240: nit: listMapKeyTagValidator and a ListMapKeyTagName const is already defined in each.go.   Let's try to maximally leverage that and avoid introducing tag parsing for that tag in this tags implementation.

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/sublistmapitem.go
```diff
@@ -0,0 +1,290 @@
...
+	}
+	return sb.String()
+}
+
+func (stv subListMapItemTagValidator) Docs() TagDoc {
+	doc := TagDoc{
+		Tag:    stv.TagName(),
+		Scopes: stv.ValidScopes().UnsortedList(),
+		Description: "Declares a validation for a subListMapItem targeting a single struct item in a slice/list with listType=map" +
+			"matching specified criteria. One of the match criteria must be on value of the listMapKey.",
```
jpbetz, line 277:
```suggestion
		Description: "Declares a validation for an item of a slice declared as a +k8s:listType=map. " +
			"The item to match is declared by providing field-value pair arguments. All +k8s:listTypeKey fields must be included in the field-value pair arguments.",
```

jpbetz: I don't think we should allow non-key field match constraints.  Requiring all key fields is already sufficient to guarantee only a single match result (unless the keys are declared wrong, which is a different topic).

But allowing non-keys significantly complicates the semantics of this tag, making it more difficult to document and adding burden to future reviewers since they need to consider not only the key lookup, but also the additional filtering and how it effects the validation.  Do we REALLY need this?

staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/tags/sublistmapitem/listcomplex/doc.go
```diff
@@ -0,0 +1,44 @@
...
+
+	NestedStruct *NestedStruct `json:"nestedStruct"`
+}
+
+type Struct struct {
+	TypeMeta int `json:"typeMeta"`
+
+	// +k8s:listType=map
+	// +k8s:listMapKey=type
+	// +k8s:subListMapItem({"type":"foo","status":"True"})=+k8s:subfield(nestedStruct)=+k8s:required
```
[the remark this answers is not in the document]
aaron-prindle: renamed to // +k8s:listMapItem

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/sublistmapitem.go
```diff
@@ -0,0 +1,290 @@
...
+	if err != nil {
+		return Validations{}, err
+	}
+
+	// This tag can apply to value and pointer fields, as well as typedefs
+	// (which should never be pointers). We need to check the concrete type.
+	t := NonPointer(NativeType(context.Type))
+	fakeComments := []string{payload}
+
+	if t.Kind != types.Slice && t.Kind != types.Array {
```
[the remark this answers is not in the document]
aaron-prindle: done

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/sublistmapitem.go
aaron-prindle: This should have been changed when the tag was split out from +k8s:subfield to +k8s:listMapItem.  This is now ScopeField.

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/sublistmapitem.go
aaron-prindle: I've updated the code to now require that the arg map keys must contain only listMapKeys and they must contain all of the listMapKeys.  This no longer allows for non-key field match constraints.  I have updated the Docs() as suggested as well related to this.

We don't really need this likely, while some code in CSR for example checks that a condition is type:[Approved|Denied|Failed] and status:True, checking just the listMapKey `type` is likely enough as only status:True is set in practice.

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/sublistmapitem.go
aaron-prindle: Fixed now.  This now supports multiple `// +k8s:listMapKey=field1`, `// +k8s:listMapKey=field2` entries.

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/sublistmapitem.go
aaron-prindle: Thanks for flagging this. Yes, all the keys should be part of the args, the code is now updated to properly account for multiple listMapKeys and this requirement

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/sublistmapitem.go
aaron-prindle: Yes, fixed now

staging/src/k8s.io/apimachinery/pkg/api/validate/sublistmapitem.go
```diff
@@ -0,0 +1,59 @@
...
+
+	"k8s.io/apimachinery/pkg/api/operation"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+)
+
+// MatchFn takes a pointer to an item and returns true if it matches the criteria.
+type MatchFn[T any] func(*T) bool
+
+// SubListMapItemByKeyValues validates a subfield of a list item where one of the selectors
+// is the listMapKey=... against a selector function.
```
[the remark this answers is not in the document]
aaron-prindle: I''ve updated the godoc to now state exactly what this function does explicitly and remove the listMapKeys mention:

```go
// ListMapItemByKeyValues finds the first item in oldList (if any) and the first
// item in newList (if any) that satisfy the 'matches' predicate. It then invokes
// 'itemValidator' on these items (if items are found).
// The fldPath passed to itemValidator is indexed to the matched item's position
// using the index from newList if a match is found there, otherwise oldList's index.
// This function processes only the *first* matching item found in each list.
// It assumes that the 'matches' predicate, targets a unique identifier (PK) and
// will match at most one element per list.
// If this assumption is violated, changes in list order can lead this function
// to have inconsistent behavior.
```

staging/src/k8s.io/apimachinery/pkg/api/validate/sublistmapitem.go
```diff
@@ -0,0 +1,59 @@
...
+	elementValidator func(ctx context.Context, op operation.Operation, fldPath *field.Path, newObj, oldObj *TItem) field.ErrorList,
+) field.ErrorList {
+	var matchedNew, matchedOld *TItem
+	var matchedIdx int
+
+	for i := range oldList {
+		if matches(&oldList[i]) {
+			matchedOld = &oldList[i]
+			matchedIdx = i
+			break
```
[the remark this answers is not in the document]
aaron-prindle: I''ve updated the godoc to now state the assumption that the matched is against a PK:

```go
...
// It assumes that the 'matches' predicate, targets a unique identifier (PK) and
// will match at most one element per list.
// If this assumption is violated, changes in list order can lead this function
// to have inconsistent behavior.
```

vendor/k8s.io/gengo/v2/comments.go
```diff
@@ -284,7 +277,6 @@ func parseTagArgs(input string) ([]string, error) {
 			}
 			return []string{string(runes[:i])}, nil
 		}
-		return nil, fmt.Errorf("unsupported character: %q", string(r))
```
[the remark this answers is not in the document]
aaron-prindle: The necessary changed is merged upstream in gengo now:
https://github.com/kubernetes/gengo/pull/296

Removing DNM

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/listmapitem.go
```diff
@@ -0,0 +1,293 @@
...
+
+func (stv listMapItemTagValidator) Docs() TagDoc {
+	doc := TagDoc{
+		Tag:    stv.TagName(),
+		Scopes: stv.ValidScopes().UnsortedList(),
+		Description: "Declares a validation for an item of a slice declared as a +k8s:listType=map." +
+			"The item to match is declared by providing field-value pair arguments. All +k8s:listMapKey fields must be included in the field-value pair arguments.",
+		Args: []TagArgDoc{
+			{
+				Description: `{"<list-map-key-field-json-name>":"<value>", "<optional-additional-field-json-name>": "<value>", ...}`,
```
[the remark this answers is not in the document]
aaron-prindle: This is fixed now.

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/listmapitem.go
```diff
@@ -0,0 +1,293 @@
...
+		jsonAnnotation, ok := tags.LookupJSON(m)
+		if ok && jsonAnnotation.Name != "" {
+			JSONName = jsonAnnotation.Name
+		}
+		if JSONName == jsonKey {
+			return m.Name, nil
+		}
+	}
+	return "", fmt.Errorf("no field with JSON name %q in type %s", jsonKey, s.Name.String())
+}
```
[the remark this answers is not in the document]
aaron-prindle: Yes, this function is not necessary and is removed now.  Use common.sh getMemberByJSON and then member.Name is equivalent.

staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/tags/listmapitem/simple/doc_test.go
```diff
@@ -0,0 +1,71 @@
...
+	}).ExpectValidateFalseByPath(map[string][]string{
+		`singleKeyItems[0]`: {singleKeyItemsMsg},
+	})
+
+	st.Value(&Struct{
+		SingleKeyItems: []Item{{Key: "NotTarget"}},
+	}).ExpectValid()
+
+	st.Value(&Struct{
+		MultipleKeyItems: []Item{{Key: "Target", Key2: "Target"}},
```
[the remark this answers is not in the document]
aaron-prindle: This now uses key:Target, key2:Target2

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/listmapitem.go
```diff
@@ -0,0 +1,293 @@
...
+		// TODO(aaron-prindle) support additional builtin types
+		var condition string
+		if fieldMember.Type.Kind == types.Pointer && fieldMember.Type.Elem != nil &&
+			fieldMember.Type.Elem.Name.Name == "string" {
+			condition = fmt.Sprintf("(item.%s != nil && *item.%s == %q)", fieldname, fieldname, listElems[jsonKey])
+		} else if fieldMember.Type.Name.Name == "string" {
+			condition = fmt.Sprintf("item.%s == %q", fieldname, listElems[jsonKey])
+		} else {
+			return FunctionLiteral{}, fmt.Errorf("must be a string or *string, but got type %s", fieldMember.Type.String())
+		}
```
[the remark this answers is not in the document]
aaron-prindle: This is simplified given that pointer fields are not supported prior and is more similar to each.go now.

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/listmapitem.go
```diff
@@ -0,0 +1,293 @@
...
+
+func createMatchFn(elemT *types.Type, listElems map[string]string) (FunctionLiteral, error) {
+	var matchFuncBody strings.Builder
+	matchFuncBody.WriteString("if item == nil { return false }\n")
+
+	var conditions []string
+	sortedKeys := make([]string, 0, len(listElems))
+	for k := range listElems {
+		sortedKeys = append(sortedKeys, k)
+	}
```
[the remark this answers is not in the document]
aaron-prindle: I've updated the validator to now use [][2]string instead of a map:
```
type Struct struct {
	TypeMeta int `json:"typeMeta"`
	// +k8s:listType=map
	// +k8s:listMapKey=key
	// +k8s:listMapItem(`[["key","Target"]]`)=+k8s:validateFalse="listMapItem SingleKeyItems[key=Target]"
	SingleKeyItems []Item `json:"singleKeyItems"`
	// +k8s:listType=map
	// +k8s:listMapKey=key
	// +k8s:listMapKey=key2
	// +k8s:listMapItem(`[["key","Target"],["key2","Target2"]]`)=+k8s:validateFalse="listMapItem MultipleKeyItems[key=Target,key2=Target2]"
	MultipleKeyItems []Item `json:"multipleKeyItems"`
}
```

staging/src/k8s.io/apimachinery/pkg/api/validate/util/util.go
aaron-prindle: I've updated listMapItem to use a [][2]string as input now instead of a map:

type Struct struct {
	TypeMeta int `json:"typeMeta"`
	// +k8s:listType=map
	// +k8s:listMapKey=key
	// +k8s:listMapItem(`[["key","Target"]]`)=+k8s:validateFalse="listMapItem SingleKeyItems[key=Target]"
	SingleKeyItems []Item `json:"singleKeyItems"`
	// +k8s:listType=map
	// +k8s:listMapKey=key
	// +k8s:listMapKey=key2
	// +k8s:listMapItem(`[["key","Target"],["key2","Target2"]]`)=+k8s:validateFalse="listMapItem MultipleKeyItems[key=Target,key2=Target2]"
	MultipleKeyItems []Item `json:"multipleKeyItems"`
}

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/listmapitem.go
```diff
@@ -0,0 +1,293 @@
...
+	matchFuncBody.WriteString(fmt.Sprintf("return %s", strings.Join(conditions, " && ")))
+	return FunctionLiteral{
+		Parameters: []ParamResult{{"item", types.PointerTo(elemT)}},
+		Results:    []ParamResult{{"", types.Bool}},
+		Body:       matchFuncBody.String(),
+	}, nil
+}
+
+func getMember(s *types.Type, fieldname string) (types.Member, error) {
+	// Assumes 's' is non-pointer struct.
```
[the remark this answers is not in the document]
aaron-prindle: This method is removed now, using getMemberFromJSON from common.sh for these cases now.

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/sublistmapitem.go
aaron-prindle: I have updated the code here to better leverage listMapKeyTagValidator.  Currently I believe the best way to do this is have listMapTagItemValidator be initialized with the validators in each.go so that `byFieldPath` can be shared with it as it is with other each.go validators.

```go
func init() {
	// Lists with list-map semantics are comprised of multiple tags, which need
	// to share information between them.
	shared := map[string]*listMetadata{} // keyed by the fieldpath
	RegisterTagValidator(listTypeTagValidator{byFieldPath: shared})
	RegisterTagValidator(listMapKeyTagValidator{byFieldPath: shared})
	RegisterTagValidator(&listMapItemTagValidator{byFieldPath: shared})

	globalEachVal = &eachValTagValidator{byFieldPath: shared, validator: nil}
	RegisterTagValidator(globalEachVal)

	globalEachKey = &eachKeyTagValidator{validator: nil}
	RegisterTagValidator(globalEachKey)
}
```

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/listmapitem.go
```diff
@@ -0,0 +1,293 @@
...
+			JSONName = jsonAnnotation.Name
+		}
+		if JSONName == jsonKey {
+			return m.Name, nil
+		}
+	}
+	return "", fmt.Errorf("no field with JSON name %q in type %s", jsonKey, s.Name.String())
+}
+
+func parseListMapKeys(commentLines []string) []string {
```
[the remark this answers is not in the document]
aaron-prindle: done

staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/tags/listmapitem/simple/doc.go
```diff
@@ -0,0 +1,44 @@
...
+	// +k8s:listMapKey=key
+	// +k8s:listMapKey=key2
+	// +k8s:listMapItem({"key":"Target","key2":"Target"})=+k8s:validateFalse="listMapItem MultipleKeyItems[key=Target,key2=Target]"
+	MultipleKeyItems []Item `json:"multipleKeyItems"`
+}
+
+type Item struct {
+	Key  string `json:"key"`
+	Key2 string `json:"key2"`
+}
```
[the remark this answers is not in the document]
aaron-prindle: I've updated the tests here for improved coverage across three test directories:
- simple
- empty
- transitions

This tests all supported usages for +k8s:listMapItem currently.

The tests do not cover the below:
- +k8s:optional does not make sense to chain
+k8s:listMapItem(...)=+k8s:optional
Panics if you directly chain during generation time as context.Member is nil
```
// hasZeroDefault returns whether the field has a default value and whether
// that default value is the zero value for the field's type.
func (rtv requirednessTagValidator) hasZeroDefault(context Context) (bool, bool, error) {
	t := NonPointer(NativeType(context.Type))
	// This validator only applies to fields, so Member must be valid.
	tagsByName, err := gengo.ExtractFunctionStyleCommentTags("+", []string{defaultTagName}, context.Member.CommentLines)

```
- +k8s:required does not make sense to chain.  If diretcly chained errors at generation time with error:
aprindle@aprindle-ssd ~/validation-gen  [aprindle-subfield-list-elem-selector-support]hack/update-codegen.sh validation
+++ [0527 07:32:09] validation: 80 targets
F0527 07:32:11.984174 2695607 targets.go:314] failed to generate validations: field k8s.io/code-generator/cmd/validation-gen/output_tests/tags/listmapitem/simple.Struct.SingleKey: tag "k8s:listMapItem": tag "k8s:required": non-pointer structs cannot use the "k8s:required" tag
!!! [0527 07:32:12] Call tree:
!!! [0527 07:32:12]  1: hack/update-codegen.sh:1132 codegen::validation(...)

- +k8s:listMapItem=+k8s:subfield(...)=+k8s:listMapItem currently does not work as a nil Context is supplied from +k8s:subfield. Errors at generation time in validator/listmapitem.go with error:  "unexpected nil context”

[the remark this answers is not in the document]
aaron-prindle: My current plan is to fix such Context and Member issues in a followup PR that addresses using +k8s:listMapItem with +k8s:unionMember

jpbetz:
Per offline discussion, I think I'm OK with this once we have:
- guards to that return sane error messages for unsupported cases (lists or pointers and such)
- test cases that clearly demonstrate the range of supported cases
- (and no cases that don't fit into one of the above)

Issue #62, opened 2025-05-15
listMap types are not validated to be unique by their key (by us)


thockin:
Should the validation code we emit include logic for that?  We do validate that items in a list-set are unique.

@jpbetz @aaron-prindle

thockin:
Played with this a bit and it's not hard BUT it turned up a fun bit of ugly.  When we have a listType=set with a struct elemtnt type, or listType=map, the error value renders as whatever Go renders, with Go-style field names. 

Following that to field.ErrorBody() we find:


```
              // TODO: internal types have panic guards against json.Marshalling to prevent  
              // accidental use of internal types in external serialized form.  For now, use  
              // %#v, although it would be better to show a more expressive output in the future  
              s = fmt.Sprintf("%s: %#v", e.Type, value) 
```

blech.

jpbetz:
Oh yuck.  I didn't realize the errors were so poor.

Yes, I agree we shouldn't be printing the whole value.

jpbetz:
Will we need to introduce a new `+k8s:listType="map"` validator?  If we validate off of the existing `+listType=map` my concern is that we've end up validating something like Pod.spec.container.envVars that doesn't have uniqueness enforced today..

thockin:
That is a good point, too.

Make it default-on but opt-out per field?

PR #70, opened 2025-05-20
Implement listmap keys uniqueness

yongruilin:
- Added UniqueByFunc to validate uniqueness of list items using a custom comparison function.
- Introduced test cases for UniqueByFunc in each_test.go to ensure correct functionality.
- Updated existing validation logic to utilize UniqueByFunc for various struct fields in generated validation files.
- Enhanced the validation framework to support early tag validation for improved processing order.

ref: https://github.com/jpbetz/validation-gen/issues/62

### DIFF
--- a/staging/src/k8s.io/apimachinery/pkg/api/validate/each.go
+++ b/staging/src/k8s.io/apimachinery/pkg/api/validate/each.go
@@ -119,6 +119,13 @@ func UniqueByReflect[T any](_ context.Context, op operation.Operation, fldPath *
 	return unique(fldPath, newSlice, SemanticDeepEqual)
 }
 
+// UniqueByFunc verifies that each element in the list map is unique by using the provided comparison function.
+// The comparison function (cmp) determines when two elements are considered equal, and if any elements
+// are found to be equal according to this function, they are reported as duplicates.
+func UniqueByFunc[T any](_ context.Context, op operation.Operation, fldPath *field.Path, newSlice, _ []T, cmp CompareFunc[T]) field.ErrorList {
+	return unique(fldPath, newSlice, cmp)
+}
+
 // unique compares every element of the slice with every other element and
 // returns errors for non-unique items.
 func unique[T any](fldPath *field.Path, slice []T, cmp func(T, T) bool) field.ErrorList {
--- a/staging/src/k8s.io/apimachinery/pkg/api/validate/each_test.go
+++ b/staging/src/k8s.io/apimachinery/pkg/api/validate/each_test.go
@@ -323,3 +323,27 @@ func testUniqueByReflect[T any](t *testing.T, name string, input []T, wantErrs i
 		}
 	})
 }
+
+type listItem struct {
+	key1 string
+	key2 string
+	val  string
+}
+
+func TestUniqueByFunc(t *testing.T) {
+	testUniqueByFunc(t, "valid", []listItem{{"key1", "key2", "val1"}, {"key1", "key2", "val2"}, {"key1", "key3", "val3"}}, 1)
+	testUniqueByFunc(t, "valid", []listItem{{"key1", "key2", "val1"}, {"key1", "key2", "val2"}, {"key1", "key2", "val3"}}, 2)
+	testUniqueByFunc(t, "valid", []listItem{{"key1", "key2", "val1"}, {"key3", "key4", "val2"}, {"key5", "key6", "val3"}}, 0)
+}
+
+func testUniqueByFunc(t *testing.T, name string, input []listItem, wantErrs int) {
+	t.Helper()
+	t.Run(name, func(t *testing.T) {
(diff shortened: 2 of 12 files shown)


staging/src/k8s.io/apimachinery/pkg/api/validate/each_test.go
```diff
@@ -222,3 +222,27 @@ func testUniqueByReflect[T any](t *testing.T, name string, input []T, wantErrs i
...
 }
+
+type listItem struct {
+	key1 string
+	key2 string
+	val  string
+}
+
+func TestUniqueByFunc(t *testing.T) {
+	testUniqueByFunc(t, "valid", []listItem{{"key1", "key2", "val1"}, {"key1", "key2", "val2"}, {"key1", "key3", "val3"}}, 1)
```
aaron-prindle, line 233: nit: Currently there are duplicate tests with the name "valid", can make the name more descriptive to explain what specifically is being tested (that or drop the test name field)

staging/src/k8s.io/apimachinery/pkg/api/validate/each_test.go
```diff
@@ -222,3 +222,27 @@ func testUniqueByReflect[T any](t *testing.T, name string, input []T, wantErrs i
...
+	testUniqueByFunc(t, "valid", []listItem{{"key1", "key2", "val1"}, {"key1", "key2", "val2"}, {"key1", "key2", "val3"}}, 2)
+	testUniqueByFunc(t, "valid", []listItem{{"key1", "key2", "val1"}, {"key3", "key4", "val2"}, {"key5", "key6", "val3"}}, 0)
```
aaron-prindle, line 235: Can we add tests for empty lists as well?  Single item lists might also make sense here.

staging/src/k8s.io/apimachinery/pkg/api/validate/each.go
```diff
@@ -102,6 +102,13 @@ func UniqueByReflect[T any](_ context.Context, op operation.Operation, fldPath *
 	return unique(fldPath, newSlice, SemanticDeepEqual)
 }
 
+// UniqueByFunc verifies that each element in the list map is unique by using the provided comparison function.
+// The comparison function (cmp) determines when two elements are considered equal, and if any elements
+// are found to be equal according to this function, they are reported as duplicates.
+func UniqueByFunc[T any](_ context.Context, op operation.Operation, fldPath *field.Path, newSlice, _ []T, cmp CompareFunc[T]) field.ErrorList {
+	return unique(fldPath, newSlice, cmp)
```
aaron-prindle, line 109: Q: Is there any ratcheting consideration necessary for if a listMapKey=... is changed or an additional key is added to a k8s object over time? Not sure if this is allowable but might make sense to note in KEP or a comment somewhere that it was considered.

thockin:
I have a sketch of how this can work, will try to post it today

thockin (commented):
I came at this a different way but headed in the same direction.  Rather than make "early" to match "late", I defined "FieldValidators" which run on every field *after* all tags.

That means we would have TypeValidator, TagValidator, and FieldValidator which makes sense to me.

https://github.com/jpbetz/validation-gen/compare/main...thockin:validation-gen:vg_ensure_unique_listMap_keys?expand=1

Mine is WIP because I didn't like how it produces errors for non-unique keys -- it logs the whole type in `%#v` syntax (gross).  I was trying to make something that looks nicer, but got distracted.

Do you want to keep running with this one?

thockin:
Here's my thinking, tell me if you disagree.

1) I hate `LateTagValidator` - it always felt like a hack.  Importantly if we do an Early version, it's possible for a validator to claim to be BOTH early and late.

2) We already have `TypeValidator` which runs on every type.  `FieldValidator` felt consistent.

3) The only existing user of `LateTagValidator` should probably be a `FieldValidator`.

I'm worried that we have built a dependency graph without putting a lot of thought into it.  Should we instead do something like this?
* each TagValidator has a method which returns a list of other tag names that it depends on
* in the sorting function, use that method to build a dependency-ordered list
* detect cycles and error-out

Pretty complicated, IMO.

Maybe we can write down rules about ordering:
- Tag validators always come first, for both types and fields.
- Type and Field validators always run after any Tag validators, and after the type or field has been processed.
- Tag validators MUST NOT depend on other Tag validators having run already.  Even when there are several tags that "work together", users might specify tags in the wrong order. Instead, they can accumulate information internally and use a Type or Field validator to finish the job.
- Type and Field validators MUST NOT depend on other Type or Field validators having run already.

thockin:
I updated https://github.com/jpbetz/validation-gen/compare/main...thockin:validation-gen:vg_ensure_unique_listMap_keys?expand=1 a little - the last commit is starting to think about how to produce better errors, but not strictly required.

thockin:
https://github.com/jpbetz/validation-gen/pull/97

Agree we can keep LateTagValidator for now and fix it when we have time - it's not urgent.  #95

PR #71, opened 2025-05-21
feat: default ratcheting for list

yongruilin:
- Add one more validate function for directly comparable input, so that the correlated elements can be directly compared by "==" instead of DeepEqual.
- If the cmp func to lookup old element is also an equivalence check, we can skip ratcheting equivalence check for update validation ratcheting.


staging/src/k8s.io/apimachinery/pkg/api/validate/each.go
```diff
@@ -33,13 +33,41 @@ type CompareFunc[T any] func(T, T) bool
 // corresponding value in oldSlice.  The value-type of the slices is assumed to
 // not be nilable.
 func EachSliceVal[T any](ctx context.Context, op operation.Operation, fldPath *field.Path, newSlice, oldSlice []T,
```
aaron-prindle, line 35: nit: Currently EachSliceVal and EachSliceValComparable are identical IIUC except for one function call (SemanticDeepEqual vs DirectEqual). Might make sense to dedupe the common logic:

func EachSliceVal[T any](...) field.ErrorList {
    return eachSliceValImpl(ctx, op, fldPath, newSlice, oldSlice, cmp, 
        isEquivalenceCompare, validator, SemanticDeepEqual[T])
}

func EachSliceValComparable[T comparable](...) field.ErrorList {
    return eachSliceValImpl(ctx, op, fldPath, newSlice, oldSlice, cmp, 
        isEquivalenceCompare, validator, DirectEqual[T])
}

staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/ratcheting/list/doc.go
```diff
@@ -0,0 +1,76 @@
+/*
+Copyright 2024 The Kubernetes Authors.
```
aaron-prindle, line 2: nit: s/2024/2025

staging/src/k8s.io/apimachinery/pkg/api/validate/each.go
```diff
@@ -33,13 +33,41 @@ type CompareFunc[T any] func(T, T) bool
...
-	cmp CompareFunc[T], validator ValidateFunc[*T]) field.ErrorList {
+	cmp CompareFunc[T], isEquivalenceCompare bool, validator ValidateFunc[*T]) field.ErrorList {
```
aaron-prindle, line 36: minor nit: isEquivalenceCompare is somewhat hard to parse what the bool does.  Perhaps skipRatchetingCheck or skipEqualityCheck?

staging/src/k8s.io/apimachinery/pkg/api/validate/each_test.go
```diff
@@ -96,13 +113,120 @@ func testEachSliceValUpdate[T any](t *testing.T, name string, input []T) {
...
+			return a.Key == b.Key
+		}),
+		false,
+	)
+	testEachSliceValComparableRatcheting(t, "same data different order", []int{11, 13, 12}, []int{11, 12, 13}, DirectEqual, true)
+	testEachSliceValComparableRatcheting(t, "same data different order", []string{"a", "c", "b"}, []string{"a", "b", "c"}, DirectEqual, true)
+	testEachSliceValComparableRatcheting(t, "less data in new, not exist in old", []string{"a", "c", "b"}, []string{"b", "c"}, DirectEqual, true)
+	testEachSliceValComparableRatcheting(t, "same data different order", []TestStruct{{11, "a"}, {13, "c"}, {12, "b"}}, []TestStruct{{11, "a"}, {12, "b"}, {13, "c"}}, DirectEqual, true)
+	testEachSliceValComparableRatcheting(t, "less data in new, not exist in old", []TestStruct{{11, "a"}, {13, "c"}, {12, "b"}}, []TestStruct{{12, "b"}, {13, "c"}}, DirectEqual, true)
+	testEachSliceValComparableRatcheting(t, "same data different order with key", []TestStructWithKey{
```
aaron-prindle, line 185: Make sense to add a test case where isEquivalenceCompare = false but the values have changed

staging/src/k8s.io/apimachinery/pkg/api/validate/each_test.go
```diff
@@ -96,13 +113,120 @@ func testEachSliceValUpdate[T any](t *testing.T, name string, input []T) {
...
+		[]NonComparableStructWithKey{
+			{Key: "a", I: 11, S: []string{"a"}}, {Key: "b", I: 12, S: []string{"b"}}, {Key: "c", I: 13, S: []string{"c"}},
+		},
+		[]NonComparableStructWithKey{
+			{Key: "a", I: 11, S: []string{"a"}}, {Key: "b", I: 12, S: []string{"b"}}, {Key: "c", I: 13, S: []string{"c"}},
+		},
+		CompareFunc[NonComparableStructWithKey](func(a, b NonComparableStructWithKey) bool {
+			return a.Key == b.Key
+		}),
+		false,
```
aaron-prindle, line 178: IIUC these are in the same order but the test name says same data different order.  Might sense to add a test case where isEquivalenceCompare = false but the values have changed here

staging/src/k8s.io/apimachinery/pkg/api/validate/each_test.go
```diff
@@ -96,13 +113,120 @@ func testEachSliceValUpdate[T any](t *testing.T, name string, input []T) {
...
+		[]NonComparableStruct{
+			{11, []string{"a"}}, {12, []string{"b"}}, {13, []string{"c"}},
+		},
+		[]NonComparableStruct{
+			{11, []string{"a"}}, {13, []string{"c"}},
+		},
+		SemanticDeepEqual,
+		true,
+	)
+	testEachSliceValRatcheting(t, "same data different order with key",
```
aaron-prindle, line 168: Might make sense to also add a test case where isEquivalenceCompare = false but the values have changed here (currently only test case same data different order w/ key IIUC)

staging/src/k8s.io/apimachinery/pkg/api/validate/each.go
```diff
@@ -33,13 +33,41 @@ type CompareFunc[T any] func(T, T) bool
...
+			continue
+		}
+		errs = append(errs, validator(ctx, op, fldPath.Index(i), &val, old)...)
+	}
+	return errs
+}
+
+// EachSliceValComparable is similar to EachSliceVal, but it assumes that the
+// elements of the slices are "direct comparable". It can use direct comparision to check
+// the equality of the correlated old and new values.
```
aaron-prindle, line 56: Currently the docstring does not explain the updated ratcheting logic or isEquivalenceCompare field.  Might make sense to update the text so users can understand what the updated params do by reading the docstring or explain at a high level the goals of the added params.

staging/src/k8s.io/apimachinery/pkg/api/validate/each.go
```diff
@@ -33,13 +33,41 @@ type CompareFunc[T any] func(T, T) bool
...
+		if op.Type == operation.Update && old != nil && (isEquivalenceCompare || SemanticDeepEqual(val, *old)) {
+			continue
+		}
+		errs = append(errs, validator(ctx, op, fldPath.Index(i), &val, old)...)
+	}
+	return errs
+}
+
+// EachSliceValComparable is similar to EachSliceVal, but it assumes that the
+// elements of the slices are "direct comparable". It can use direct comparision to check
```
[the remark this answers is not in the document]
yongruilin: Updated.

staging/src/k8s.io/apimachinery/pkg/api/validate/each.go
yongruilin: I changed it to `skipEquivalenceCheck`.

staging/src/k8s.io/apimachinery/pkg/api/validate/each.go
yongruilin: Updated the doc comment.

staging/src/k8s.io/apimachinery/pkg/api/validate/each.go
```diff
@@ -32,14 +32,42 @@ type CompareFunc[T any] func(T, T) bool
 // validation function.  The comparison function is used to find the
 // corresponding value in oldSlice.  The value-type of the slices is assumed to
 // not be nilable.
+//
+// The skipEquivalenceCheck parameter controls validation ratcheting behavior during updates:
```
thockin, line 36: I find this very hard to reason about.

Would it be simpler to:
* first: rename the existing "cmp" argument to "match" and document it more clearly that the semantics of the match function depend on whether the list is just a list or has map or set semantics.
* second: add a "cmp" argument (same function signature) and document that it does a full comparison for the purpose of bypassing validation on unchanged values 

Then it's not some boolean that modifies the meaning, it's just two function pointers.  Sometimes they will be the same and sometimes they won't.

yongruilin: Maybe I can expand it to 4 functions, eachValForMap/NonMap/ComparableForMap/ComparableNonMap

yongruilin: Sounds good. I've updated the accordingly PTAL.

thockin (commented):
LGTM, just small things!

thockin:
LGTM

PR #72, opened 2025-05-21
Best effort to directly compare by using IsDirectComparable

yongruilin:
Structs with all its fields comparable (not pointers), should be able to be compared directly with "==". 
- Move IsDirectComparable(T types.Type) into validators package.
- Apply on "each" tag validator.
- Apply on "immutable" tag validator.

### DIFF
--- a/staging/src/k8s.io/code-generator/cmd/validation-gen/validators/common.go
+++ b/staging/src/k8s.io/code-generator/cmd/validation-gen/validators/common.go
@@ -104,3 +104,26 @@ func rootTypeString(src, dst *types.Type) string {
 	}
 	return src.String() + " -> " + dst.String()
 }
+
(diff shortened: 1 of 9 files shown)


staging/src/k8s.io/code-generator/cmd/validation-gen/validators/common.go
```diff
@@ -104,3 +104,26 @@ func rootTypeString(src, dst *types.Type) string {
 	}
 	return src.String() + " -> " + dst.String()
 }
+
+// IsDirectComparable returns true if the type is safe to compare using "==".
+// It is similar to gengo.IsComparable, but is doesn't consider Pointers
```
aaron-prindle, line 109: nit: s/ is doesn't / it doesn't

thockin:
LGTM

PR #75, opened 2025-05-24
Followup to #55

thockin:
1) Better errors from matcher when it holds a ptr.

2) Simplify some tests for reading

3) Adopt the same utility funcs in other places, delete half-duplicates

4) Export IsNilableType the same way

5) Move the utility guncs to a pkg.


thockin:
Rebased this - would be nice to get in.

staging/src/k8s.io/apimachinery/pkg/util/validation/field/error_matcher.go
```diff
@@ -91,7 +91,16 @@ func (m ErrorMatcher) Render(e *Error) string {
 		if s, ok := e.BadValue.(string); ok {
 			buf.WriteString(fmt.Sprintf("Value=%q", s))
 		} else {
-			buf.WriteString(fmt.Sprintf("Value=%v", e.BadValue))
+			rv := reflect.ValueOf(e.BadValue)
+			if rv.Kind() == reflect.Pointer && !rv.IsNil() {
+				rv = rv.Elem()
+			}
+			if rv.CanInterface() {
+				buf.WriteString(fmt.Sprintf("Value=%v", rv.Interface()))
```
yongruilin, line 99: Is printing the interface better?

yongruilin (commented):
LGTM, just one question to clarify.

staging/src/k8s.io/apimachinery/pkg/util/validation/field/error_matcher.go
thockin: Before this change it would print something like `Value=*0`but the only way to see if it is a pointer is to go thru `reflect`.  Once you have a `reflect.Value`, the only way to get the actual value as an `any` is to call `.Interface()`

yongruilin: I see. Thanks for the explanation. Could you rebase and we merge it.

PR #79, opened 2025-05-29
validation-gen support ratcheting for map(eachKey/eachVal)

yongruilin:
- support ratcheting for map's eachVal
- support ratcheting for map's eachKey


staging/src/k8s.io/apimachinery/pkg/api/validate/each.go
```diff
@@ -102,6 +102,15 @@ func EachMapKey[K ~string, T any](ctx context.Context, op operation.Operation, f
 	validator ValidateFunc[*K]) field.ErrorList {
 	var errs field.ErrorList
 	for key := range newMap {
+		var old *K
```
aaron-prindle, line 105: The current go doc string for EachMapKey states:

// EachMapKey validates each element of newMap with the specified
// validation function.  The oldMap argument is not used.

The oldMap argument is used now IIUC so I believe the text should be updated there

staging/src/k8s.io/apimachinery/pkg/api/validate/each.go
```diff
@@ -73,17 +73,24 @@ func lookup[T any](list []T, target T, cmp func(T, T) bool) *T {
 	return nil
 }
 
-// EachMapVal validates each element of newMap with the specified validation
-// function and, if the corresponding key is found in oldMap, the old value.
-// The value-type of the slices is assumed to not be nilable.
+// EachMapVal validates each value in newMap using the specified validation
+// function, passing the corresponding old value from oldMap if the key exists in oldMap.
+// For update operations, it implements validation ratcheting by skipping validation
+// when the old value exists and the equiv function confirms the values are equivalent.
```
aaron-prindle, line 79: nit: Might make sense for the go doc or a comment around the equiv != nil line to explain that equiv can be nil and that it provides a "kill switch" for value-based ratcheting.

staging/src/k8s.io/apimachinery/pkg/api/validate/each_test.go
```diff
@@ -230,13 +230,92 @@ func testEachMapVal[T any](t *testing.T, name string, input map[string]T) {
...
 		}
-		_ = EachMapVal(context.Background(), operation.Operation{}, field.NewPath("test"), input, nil, vfn)
+		_ = EachMapVal(context.Background(), operation.Operation{}, field.NewPath("test"), input, nil, nil, vfn)
 		if calls != len(input) {
 			t.Errorf("expected %d calls, got %d", len(input), calls)
 		}
 	})
 }
 
+func TestEachMapValRatcheting(t *testing.T) {
```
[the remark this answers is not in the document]
yongruilin: I've updated the test cases.

staging/src/k8s.io/apimachinery/pkg/api/validate/each.go
yongruilin: Good catch, updated.

PR #83, opened 2025-06-04
feat: Add validation ratcheting for subfields tag

yongruilin:
Current implementation of subfield tags always get the field as "Nilable", which would require always using SemanticDeepEqual.


staging/src/k8s.io/apimachinery/pkg/api/validate/subfield.go
```diff
@@ -36,6 +36,9 @@ func Subfield[Tstruct any, Tfield any](ctx context.Context, op operation.Operati
 	if oldStruct != nil {
 		oldVal = getField(oldStruct)
 	}
+	if op.Type == operation.Update && SemanticDeepEqual(newVal, oldVal) {
```
aaron-prindle, line 39: Can you explain the subfield ratcheting rules?  IIUC a cross-field tag could be put on subfields, is it the case currently that when updating an object, if a subfield hasn't changed (determined by SemanticDeepEqual), the validation is skipped entirely?  This aligns with the ratcheting principles outlined in the KEP IIUC but is this correct for subfield + cross-field?

I am not suggesting we necessarily need to change the logic here but I just wanted to understand:
- Are the below practical cases? 
- Do we need to prevent them from being possible given this logic?

```go
+k8s:subfield(foo)=+k8s:unionMember

+k8s:subfield(foo)=+k8s:minimum(siblingField) // currently doesn't exist
```

thockin: This supports the argument for putting cross-field validations on the parent?

For a simple case of `field X < field Y`, we need to check both participating fields, not just the one we attach to.
 
For subfields we a) need a way to express `subfield(X) < subfield(Y)`; and b) need to ratchet-check both, I think?

thockin, line 39: we could, again, pass an equiv function in to subfield so we can use direct compare.

Maybe just put a TODO here?  It doesn't seem critical for now

thockin (commented):
LGTM

staging/src/k8s.io/apimachinery/pkg/api/validate/subfield.go
yongruilin: I tried to implement this but gave up. Because Subfield's GetFieldFunc always output an Nilable, I need to do some special handling e.g. check if it is a pointer and deref it in runtime, etc.

Anyway, I dropped a TODO for now.

PR #89, opened 2025-06-09
chore: Add validation error for unsupported map of slices in validation-gen

yongruilin:
Block validations attached on map of slices.
Follow up of https://github.com/jpbetz/validation-gen/commit/d2f7fee5a3a5ac15791f527d74bd3aa4d5f54e8e


staging/src/k8s.io/code-generator/cmd/validation-gen/validation.go
```diff
@@ -407,6 +407,9 @@ func (td *typeDiscoverer) discoverType(t *types.Type, fldPath *field.Path) (*typ
 		} else if validations.Empty() {
 			klog.V(6).InfoS("no type-attached validations", "type", t)
 		} else {
+			if util.NonPointer(util.NativeType(t)).Kind == types.Map && util.NonPointer(util.NativeType(t)).Elem.Kind == types.Slice {
```
aaron-prindle, line 410: Can we add a test case for these two cases in validation_test.go to make sure they are firing in the cases we  expect?

PR #97, opened 2025-06-12
Ensure unique list map keys

### LINKED ISSUE #62: listMap types are not validated to be unique by their key (by us)

thockin:
Closes #62 
Closes #70

We will probably need to (eventually) add a way for some fields to say "I am a listmap but might be non-unique" - as screwed up as that is.


thockin:
Comments addressed

aaron-prindle:
LGTM

PR #100, opened 2025-06-16
feat: Add validation ratcheting for Union

yongruilin:


staging/src/k8s.io/apimachinery/pkg/api/validate/union.go
```diff
@@ -101,13 +112,21 @@ func DiscriminatedUnion[T any, D ~string](_ context.Context, op operation.Operat
 					len(extractorFns), len(union.members))),
 		}
 	}
-
+	var changed bool
 	discriminatorValue := discriminatorExtractor(obj)
+	if op.Type == operation.Update && !changed {
```
aaron-prindle, line 117: `!changed` is always true here IIUC as it is just declared prior, it might make sense to remove it

staging/src/k8s.io/apimachinery/pkg/api/validate/union.go
```diff
@@ -101,13 +112,21 @@ func DiscriminatedUnion[T any, D ~string](_ context.Context, op operation.Operat
...
+		oldDiscriminatorValue := discriminatorExtractor(oldObj)
+		changed = discriminatorValue != oldDiscriminatorValue
+	}
 
 	for i, extractor := range extractorFns {
 		member := union.members[i]
 		isDiscriminatedMember := string(discriminatorValue) == member.discriminatorValue
 		fieldValue := extractor(obj)
+		if op.Type == operation.Update && !changed {
```
aaron-prindle, line 126: do we need nil protection here? I think that if oldObj is nil and we call extractor(oldObj) we could get a panic.  Might want to add a test case for this as well

`if op.Type == operation.Update && oldObj != nil && !changed {`

staging/src/k8s.io/apimachinery/pkg/api/validate/union_test.go
```diff
@@ -80,16 +80,24 @@ func TestUnion(t *testing.T) {
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
+			// Create mock extractors that return predefined values instead of
```
aaron-prindle, line 83: The current mock extractors can't distinguish between old and new objects, so they always return identical values, making `changed` always false IIUC.  I believe for ratcheting the test should be updated to use real extractors:

```
extractors := []ExtractorFn[*TestStruct, any]{
    func(obj *TestStruct) any { return obj.A },
    func(obj *TestStruct) any { return obj.B },
    func(obj *TestStruct) any { return obj.C },
}
```

staging/src/k8s.io/apimachinery/pkg/api/validate/union.go
```diff
@@ -53,13 +53,24 @@ func Union[T any](_ context.Context, op operation.Operation, fldPath *field.Path
 		}
 	}
 	var specifiedFields []string
+	var changed bool
 	for i, extractor := range extractorFns {
 		fieldValue := extractor(obj)
+		if op.Type == operation.Update && !changed {
```
aaron-prindle, line 59: This check is odd to me here:
		if op.Type == operation.Update && !changed {
		
and then later we have:

	// If all union members are unchanged, we don't need to validate the union.
	if op.Type == operation.Update && !changed {
		return nil
	}
	
	Can we short circuit here?
	
	Also not sure what the common case is but if it is that the union doesn't change then each time this is called we build specifiedFields but unnecessarily if we hit:
	
		if op.Type == operation.Update && !changed {
		return nil
	}

staging/src/k8s.io/apimachinery/pkg/api/validate/union.go
yongruilin: Another thought, should we handle the nil check inside the `extractor` ?

staging/src/k8s.io/apimachinery/pkg/api/validate/union.go
yongruilin: > Can we short circuit in the original loop?

We can only return nil when all fields are "unchanged" (!changed), so we need to iterate through all the extractors. I don't think we can "shortcircut" more. Actually, there is already some kind of short-circuit logic, as during the loop, if there is `changed`, it won't continue to do the SemanticDeepEqual
 
> might make sense to built specifiedFields after we know if a change has been made
One way to do it is to separate the ratcheting check loop with the specificedFields loop, but that would add the extra time. sth like this
```
	if op.Type == operation.Update {
		changed := false
		for _, extractor := range extractorFns {
			fieldValue := extractor(obj)
			oldFieldValue := extractor(oldObj)
			// TODO: consider passing an equiv function to DirectEqual instead of SemanticDeepEqual
			// if value can be compared directly, to improve performance.
			changed = !SemanticDeepEqual(fieldValue, oldFieldValue)
		}
		if !changed {
			return nil
		}
	}
	var specifiedFields []string
	for i, extractor := range extractorFns {
		fieldValue := extractor(obj)
		rv := reflect.ValueOf(fieldValue)
		if rv.IsValid() && !rv.IsZero() {
			specifiedFields = append(specifiedFields, union.members[i].fieldName)
		}
	}
...
```

staging/src/k8s.io/apimachinery/pkg/api/validate/union.go
yongruilin: Looks like the obj is `T` in validate.Union, not sure if we can check if it is nil there..

yongruilin: Updated.

staging/src/k8s.io/apimachinery/pkg/api/validate/union_test.go
yongruilin: Updated, added new unittest for ratcheting

staging/src/k8s.io/apimachinery/pkg/api/validate/union_test.go
```diff
@@ -80,16 +80,24 @@ func TestUnion(t *testing.T) {
...
 				extractors[i] = func(_ *testMember) any { return val }
 			}
 
-			got := Union(context.Background(), operation.Operation{Type: operation.Update}, nil, &testMember{}, nil, NewUnionMembership(tc.fields...), extractors...)
+			got := Union(context.Background(), operation.Operation{}, nil, &testMember{}, nil, NewUnionMembership(tc.fields...), extractors...)
 			if !reflect.DeepEqual(got, tc.expected) {
 				t.Errorf("got %v want %v", got, tc.expected)
 			}
+
+			// Test ratcheting, expect no error when the object is unchanged.
```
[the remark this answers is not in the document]
yongruilin: Updated.

staging/src/k8s.io/apimachinery/pkg/api/validate/union_test.go
```diff
@@ -127,16 +129,187 @@ func TestDiscriminatedUnion(t *testing.T) {
...
+	},
+	func(s *testDiscriminatedStruct) any {
+		if s != nil {
+			return s.M2
+		}
+		return nil
+	},
+}
+
+func TestDiscriminatedUnionRatcheting(t *testing.T) {
```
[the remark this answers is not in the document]
yongruilin: Added.

aaron-prindle:
LGTM

thockin:
I didn't look suuuuuuuper hard at this one (leaning on Aaron and Joe's reviews) but this looks like what I expected.

staging/src/k8s.io/apimachinery/pkg/api/validate/union_test.go
```diff
@@ -127,16 +129,202 @@ func TestDiscriminatedUnion(t *testing.T) {
...
+}
+
+func TestUnionRatcheting(t *testing.T) {
+	testCases := []struct {
+		name      string
+		oldStruct *testStruct
+		newStruct *testStruct
+		expected  field.ErrorList
+	}{
+		{
```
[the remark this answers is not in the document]
jpbetz: Yes, thanks!

jpbetz (approved):
/approve
/lgtm

PR #101, opened 2025-06-17
feat(validation-gen): add +k8s:eqOneOf tag which validates the field equals one of the payload array values

aaron-prindle:
Extends `validation-gen` to support the `+k8s:eqOneOf` tag, enabling validation that a field's value matches one of a specified set of allowed values.

- Adds new validation tag `+k8s:eqOneOf`: Validates that a field contains one of the allowed values specified in a JSON array
-- JSON array syntax: Uses JSON arrays for specifying allowed values
- Supports comparable types: Works with strings, integers (all sizes), and booleans, including typedefs

### Example Usage

```go
type Config struct {
    // +k8s:eqOneOf=`["small","medium","large"]`
    Size string `json:"size"`
    
    // +k8s:eqOneOf=`[80,443,8080]`
    Port int `json:"port"`
    
    // +k8s:eqOneOf=`[true,false]` // redundant but fleshes out comparable
    AllowRetry bool `json:"allowRetry"`
    
    // +k8s:eqOneOf=`["active","inactive","pending"]`
    Status StatusType `json:"status"`
}
```

### Why is this needed? - CSR Validation

CSR has one validations that requires using +k8s:eqOneOf, this validation here that checks the usage is one of list of ~20 valid usages:

https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/certificates/v1/types.go#L110C1-L125C80
```go
	// usages specifies a set of key usages requested in the issued certificate.
	//
	// Requests for TLS client certificates typically request: "digital signature", "key encipherment", "client auth".
	//
	// Requests for TLS serving certificates typically request: "key encipherment", "digital signature", "server auth".
	//
	// Valid values are:
	//  "signing", "digital signature", "content commitment",
	//  "key encipherment", "key agreement", "data encipherment",
	//  "cert sign", "crl sign", "encipher only", "decipher only", "any",
	//  "server auth", "client auth",
	//  "code signing", "email protection", "s/mime",
	//  "ipsec end system", "ipsec tunnel", "ipsec user",
	//  "timestamping", "ocsp signing", "microsoft sgc", "netscape sgc"
	// +listType=atomic
	Usages []KeyUsage `json:"usages,omitempty" protobuf:"bytes,5,opt,name=usages"`
```

https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/certificates/validation/validation.go#L139-L165
https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/certificates/validation/validation.go#L436-L438
```go
var (
	allValidUsages = sets.NewString(
		string(certificates.UsageSigning),
		string(certificates.UsageDigitalSignature),
		string(certificates.UsageContentCommitment),
		string(certificates.UsageKeyEncipherment),
…
```

```go
func hasUnknownUsage(usages []certificates.KeyUsage) bool {
	for _, usage := range usages {
		if !allValidUsages.Has(string(usage)) {
			return true
		}
	}
	return false
}
```

### CSR Declarative Validation Tag Usage For v1.34

```go
// CertificateSigningRequestSpec contains the certificate request.
type CertificateSigningRequestSpec struct {
        // +k8s:eqOneOf=["signing","digital signature","content commitment","key encipherment","key agreement","data encipherment","cert sign","crl sign","encipher only","decipher only","any","server auth","client auth","code signing","email protection","s/mime","ipsec end system","ipsec tunnel","ipsec user","timestamping","ocsp signing","microsoft sgc","netscape sgc"]
	Usages []KeyUsage `json:"usages,omitempty" protobuf:"bytes,5,opt,name=usages"`
```

### Notes
Currently +k8s:eqOneOf bool arrays are supported and tested to fully implement comparable but it is unlikely it would be used in practice
The name +k8s:eqOneOf is currently chosen as the idea was it would part of equality.go and there would eventually be +k8s:eq, +k8s:neq, +k8s:eqOneOf, +k8s:neqOneOf


aaron-prindle:
In actually looking at the CSR types, it seems that a KeyUsage is an enum so likely we can just use +k8s:enum here.  Closing this PR

PR #106, opened 2025-07-01
Fix go fmt and lint issues.

lalitc375:
Fix go fmt and lint issues.

* Also add github actions to prevent future regiessions.

#### What type of PR is this?

/kind cleanup


.github/workflows/validation-gen-ci-test.yml
```diff
@@ -21,6 +21,15 @@ jobs:
         hack/update-codegen.sh validation
         git diff --exit-code
 
+    - name: Run go formatter and check for diffs
```
aaron-prindle, line 24: nit: might make sense to rename this to

- name: Run gofmt and check for diffs

.github/workflows/validation-gen-ci-test.yml
```diff
@@ -21,6 +21,15 @@ jobs:
...
+      run: |
+        hack/verify-gofmt.sh
+        git diff --exit-code
+
+    - name: Run Go Lang ci lint
```
aaron-prindle, line 29: nit: might make sense to rename this to

name: Run Golang CI Lint
OR
name: Run golangci-lint CI Test

golangci-lint might usually written that way and is possibly easier to parse vs Golang CI Lint

aaron-prindle:
Great, thanks for this!  Left 2 small nits related to the GH action step names.

Additionally on the upstream PR here: https://github.com/kubernetes/kubernetes/pull/132629

I see this failure as well:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/132629/pull-kubernetes-linter-hints/1939833008307572736
```
verify: golangci-lint-pr-hints expand_less	18m45s
{ ScriptError 
staging/src/k8s.io/code-generator/cmd/validation-gen/testscheme/testscheme.go:421:4: QF1008: could remove embedded field "T" from selector (staticcheck)
	v.T.Helper()
	  ^
staging/src/k8s.io/code-generator/cmd/validation-gen/testscheme/testscheme.go:423:4: QF1008: could remove embedded field "T" from selector (staticcheck)
	v.T.Run(fmt.Sprintf("%T", v.value), func(t *testing.T) {
	  ^
staging/src/k8s.io/code-generator/cmd/validation-gen/util/util.go:70:3: QF1003: could use tagged switch on t.Kind (staticcheck)
		if t.Kind == types.Alias {
		^
staging/src/k8s.io/code-generator/cmd/validation-gen/validators/enum.go:220:9: SA1019: gengo.ExtractCommentTags is deprecated: Prefer codetags.Extract and codetags.Parse. (staticcheck)
	return gengo.ExtractCommentTags("+", t.CommentLines)[enumTagName] != nil
	       ^
}
```

Are these golang-cli-lint-pr-hints something we should also add in the PR here?

.github/workflows/validation-gen-ci-test.yml
lalitc375: done

lalitc375:
> Great, it might make sense to also add this verify check to our CI as well in this case:

I have replaced golangci-lint with verify-golangci-lint-pr-hints, As the later calls the former internally.

lalitc375:
> > Great, it might make sense to also add this verify check to our CI as well in this case:
> 
> I have replaced golangci-lint with verify-golangci-lint-pr-hints, As the later calls the former internally.

verify-golangci-lint-pr-hints can't be used in github actions, as it has deps on prow jobs envs. Reverted back to  golangci-lint

aaron-prindle:
To confirm that the changes here pass with all green upstream, can you add this commit to the PR at:
https://github.com/kubernetes/kubernetes/pull/132629

and verify we get all green?  Currently there is some CI test failures there and it might make sense to confirm we are all green upstream with the changes in this PR to avoid additional changes.

PR #105, opened 2025-06-30
[DNM] add benchmark of list ratcheting

yongruilin:
Ref: #104

```
~ go test -bench=. ./staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/_codegenignore/b
enchmark/ratcheting -benchmem

goos: linux
goarch: amd64
pkg: k8s.io/code-generator/cmd/validation-gen/output_tests/_codegenignore/benchmark/ratcheting
cpu: AMD EPYC 7B13
BenchmarkRatchetingApproaches/NoRatcheting-16               2745            463170 ns/op          116385 B/op       3900 allocs/op
BenchmarkRatchetingApproaches/MapBasedLookup-16             2384            531058 ns/op          146811 B/op       5305 allocs/op
BenchmarkRatchetingApproaches/LinearScan-16                  384           3086953 ns/op           26331 B/op       1300 allocs/op
BenchmarkChangeScenarios/LinearScan_LowChange-16             388           3114665 ns/op           55061 B/op       3100 allocs/op
BenchmarkChangeScenarios/MapBasedLookup_LowChange-16        1642            741302 ns/op          175785 B/op       7105 allocs/op
BenchmarkChangeScenarios/NoRatcheting_LowChange-16          2392            433343 ns/op          116182 B/op       3900 allocs/op
BenchmarkChangeScenarios/LinearScan_MediumChange-16          352           3162654 ns/op           69595 B/op       3300 allocs/op
BenchmarkChangeScenarios/MapBasedLookup_MediumChange-16     1537            778392 ns/op          191282 B/op       7305 allocs/op
BenchmarkChangeScenarios/NoRatcheting_MediumChange-16       2856            434013 ns/op          116249 B/op       3900 allocs/op
BenchmarkChangeScenarios/LinearScan_HighChange-16            369           3345363 ns/op           82681 B/op       3500 allocs/op
BenchmarkChangeScenarios/MapBasedLookup_HighChange-16       1548            825201 ns/op          206559 B/op       7505 allocs/op
BenchmarkChangeScenarios/NoRatcheting_HighChange-16         2626            419496 ns/op          116261 B/op       3900 allocs/op
BenchmarkChangeScenarios/LinearScan_NoChange-16              393           3159918 ns/op           48000 B/op       3000 allocs/op
BenchmarkChangeScenarios/MapBasedLookup_NoChange-16         1711            670892 ns/op          166778 B/op       7005 allocs/op
[...]

PR #107, opened 2025-07-01
No lookup correlated for non-map/set list in ratcheting

yongruilin:
ref: #105

### DIFF
--- a/staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/ratcheting/list/doc_test.go
+++ b/staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/ratcheting/list/doc_test.go
@@ -100,4 +100,11 @@ func Test_StructSlice(t *testing.T) {
 	}).ExpectMatches(field.ErrorMatcher{}.ByType().ByField(), field.ErrorList{
 		field.Invalid(field.NewPath("sliceNonComparableStructField[0]"), "", ""),
 	})
+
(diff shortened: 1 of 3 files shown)


staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/ratcheting/list/doc_test.go
```diff
@@ -65,7 +65,12 @@ func Test_StructSlice(t *testing.T) {
 		TypeDefSliceField:             MySlice{1, 2},
 		SliceStructField:              []DirectComparableStruct{{IntField: 2}, {IntField: 1}},
 		SliceNonComparableStructField: []NonDirectComparableStruct{{IntPtrField: ptr.To(2)}, {IntPtrField: ptr.To(1)}},
-	}).ExpectValid()
+	}).ExpectMatches(field.ErrorMatcher{}.ByType().ByField(), field.ErrorList{
+		field.Invalid(field.NewPath("sliceField[0]"), "", ""),
```
thockin, line 69: I think that any time we use `ExpectMatches()` and `Invalid()` we should either match on origin or match on details substring, so you could include `"field SliceField[*]"` 

But I think it might make sense to rename the fields as something like "AtomicSliceField", "MapSliceField" (you already have "Set" in the name).

thockin (commented):
I agree with this - have got David and Jordan to agree?

PR #111, opened 2025-07-08
chore: Removed unnecessary type argument

yongruilin:
- Remove unnecessary type assertion when they can be inferred.
  - For VariableGen, only requires when initialization func has empty args
  - For FunctionGen, it would always pass in an "obj" with type asserted.

### DIFF
--- a/staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/type_args/zz_generated.validations.go
+++ b/staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/type_args/zz_generated.validations.go
@@ -70,7 +70,7 @@ func Validate_T1(ctx context.Context, op operation.Operation, fldPath *field.Pat
 				return nil
 			}
 			// call field-attached validations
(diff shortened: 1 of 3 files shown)


staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/type_args/zz_generated.validations.go
```diff
@@ -67,7 +67,7 @@ func Validate_T1(ctx context.Context, op operation.Operation, fldPath *field.Pat
 			if op.Type == operation.Update && (obj == oldObj || (obj != nil && oldObj != nil && *obj == *oldObj)) {
 				return nil // no changes
 			}
-			errs = append(errs, validate.FixedResult[*primitives.T1](ctx, op, fldPath, obj, oldObj, false, "T1.S1")...)
```
thockin, line 70: The point of this test was to prove that a validator with explicit type args would render correctly.  If we think we NEVER need type args anymore (why did we need them at all?  @jpbetz) then we should nuke this whole test

thockin:
See #176 instead

PR #112, opened 2025-07-08
Change union ratcheting to only look at membership

thockin:
We don't care if the values of the members changed, only which members are set / unset.


yongruilin (commented):
LGTM.
I'll leave merge to @aaron-prindle since he originally added the extractorFn.

aaron-prindle:
LGTM, merging

PR #114, opened 2025-07-09
Update validation-gen/generate_docs.sh

yongruilin:
Keep it consistent with https://github.com/kubernetes/kubernetes/pull/130868


thockin:
How about `--docs=(json|md)` with JSON being default?  Or even add a non-JSON format that is easier to look at on a terminal?

PR #115, opened 2025-07-09
chore: fix forward on validation versioned test

yongruilin:
Keep consistent with https://github.com/kubernetes/kubernetes/pull/132465
This would pave the road for the following migration of CSR.


thockin (commented):
Thanks!

/lgtm
/approve

Issue #87, opened 2025-06-06
TODO: Clean up tests


thockin:
All users of `.ExpectInvalid()` would probably be better with `ExpectMatches()` or `ExpectValidateFalseByPath()`.  Ditto for `ExpectRegexpsByPath()`

thockin:
Copying for the record:

Related comment:
https://github.com/jpbetz/validation-gen/pull/110#discussion_r2201768989

For tests, the only 3 Expect we should use are:

- `ExpectValid()`
- `ExpectValidateFalseByPath()` (for tests that are basically tracing execution with validateFalse)
- `ExpectMatches()` (for anything more than just validateFalse)

All the rest should be converted and EOL'ed

PR #119, opened 2025-07-10
feat: add +k8s:zeroOrOneOfMember tag which uses shared union methods

aaron-prindle:
This PR adds new tag: `+k8s:zeroOrOneOfMember` to validation gen.  This tag is similar to +k8s:unionMember only it differs in the that the union is allowed to be empty (vs +k8s:unionMember which requires one member).  The PR does the following:

- refactors union.go code to be resusable for additional union style tags
- adds new zerorone.go code which leverages the refactored union code to create the +k8s:zeroOrOneOfMember tag
- adds new tests for +k8s:zeroOrOneOfMember tags based off of +k8s:unionMember (undiscriminated) tests only with the tests modified to use the new tags and expect the modified semantics


staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/tags/zerooroneof/custom_members/doc.go
```diff
@@ -0,0 +1,44 @@
...
+	NonUnionField string `json:"nonUnionField"`
+
+	// +k8s:zeroOrOneOfMember(memberName: "CustomM1")
+	// +k8s:optional
+	M1 *M1 `json:"m1"`
+
+	// +k8s:zeroOrOneOfMember(memberName: "CustomM2")
+	// +k8s:optional
+	M2 *M2 `json:"m2"`
+}
```
jpbetz, line 40: Let's also test:

- With a discriminator (optional I assume)
- Two zeroOrOneOf's in the same struct
- A zeroOrOneOf and a union in the same struct (do names conflict?)
- an empty zeroOrOneOf (no members, but has a discriminator)

staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/tags/zerooroneof/custom_members/doc.go
```diff
@@ -0,0 +1,44 @@
...
+
+var localSchemeBuilder = testscheme.New()
+
+// Non-discriminated zero-or-one-of union with custom member names
+type Struct struct {
+	TypeMeta int
+
+	NonUnionField string `json:"nonUnionField"`
+
+	// +k8s:zeroOrOneOfMember(memberName: "CustomM1")
```
jpbetz, line 33: Are we all okay with calling this "zeroOrOneOf"?  It's quite a a mouth full.  But it's also clear to me what it's suppose to do.  I just wish the naming was more consistent with union.  We could have "oneOf" and "zeroOrOneOf".  Alternative we could have "union" and "optionalUnion" or something, but I don't like "optionalUnion" at all..

Thoughts?

@thockin @yongruilin

jpbetz: Another way we could do this would be to add support for explicitly declaring unions on the parent via a `+k8s:union` tag like so:

```
// +k8s:union(optional: true)
type Struct struct {
	TypeMeta int

	// +k8s:unionMember
	// +k8s:optional
	M1 *M1 `json:"m1"`

	// +k8s:unionMember
	// +k8s:optional
	M2 *M2 `json:"m2"`
}
```

This would provide a way to parameterize unions (both named and unnamed) and could make it possible to express the "zeroOr" part of the "zeroOrOneOf" tag using the existing union tag..

staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/tags/zerooroneof/custom_members/doc.go
jpbetz: I think I'm fine with "let's not support the discriminator for zeroOrOneOf until we know we need it".

staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/tags/zerooroneof/custom_members/doc.go
jpbetz: OK, so decision is we stick with `zeroOrOneOfMember` for now but leave the door to revisit post 1.34 code freeze.

aaron-prindle: Sg, we can stick with +k8s:zerOrOneOfMember for v1.34 and then revisit the discussion here

jpbetz (approved):
LGTM

Test coverage looks sufficient now.  Let's merge and iterate post-1.34

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/union.go
```diff
@@ -32,6 +32,7 @@ var unionValidator = types.Name{Package: libValidationPkg, Name: "Union"}
 
 var newDiscriminatedUnionMembership = types.Name{Package: libValidationPkg, Name: "NewDiscriminatedUnionMembership"}
 var newUnionMembership = types.Name{Package: libValidationPkg, Name: "NewUnionMembership"}
+var unionVariablePrefix = "unionMembershipFor"
```
thockin, line 35: Could have been `const` :)

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/union.go
```diff
@@ -305,3 +201,136 @@ func (us unions) getOrCreate(name string) *union {
...
+	}
+	nt := util.NativeType(member.Type)
+	switch nt.Kind {
+	case types.Pointer, types.Map, types.Slice:
+		extractor.Body = fmt.Sprintf("if obj == nil {return false}; return obj.%s != nil", member.Name)
+	case types.Builtin:
+		extractor.Body = fmt.Sprintf("if obj == nil {return false}; var z %s; return obj.%s != z", member.Type, member.Name)
+	default:
+		// This should be caught before we get here, but JIC.
+		extractor.Body = fmt.Sprintf("if obj == nil {return false}; return false /* unsupported union member kind: %s */", nt.Kind)
```
thockin, line 273: This used to be an error returned to the tool, causing generation to fail.  Sticking it into the body is far less obvious and not likely to be seen, should it occur.

Couldn't we return error here?

staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/tags/zerooroneof/simple/doc_test.go
```diff
@@ -0,0 +1,54 @@
...
+
+	// Empty union is valid for zeroOrOneOf
+	st.Value(&Struct{}).ExpectValid()
+
+	st.Value(&Struct{M1: &M1{}}).ExpectValid()
+	st.Value(&Struct{M2: &M2{}}).ExpectValid()
+	st.Value(&Struct{M3: "a string"}).ExpectValid()
+	st.Value(&Struct{M4: ptr.To("a string")}).ExpectValid()
+
+	st.Value(&Struct{M1: &M1{}, M2: &M2{}}).ExpectInvalid(
```
thockin, line 37: https://github.com/jpbetz/validation-gen/issues/87

thockin (commented):
This is fine to me, the only thing I care about is the "should never happen" error being returned vs emitted in code

Issue #113, opened 2025-07-09
Improve testing - add unit tests for .../validation-gen/validators/*


thockin:
We have relied on output tests to check the validators, but those are no good for testing error cases.

We should probably write some simple test scaffolding to call `GetValidations()` on each validator, PARTICULARLY to hit the various expected error cases (e.g. "plain struct can't be optional").

thockin:
I thought of #92 as "run the tool against bad inputs and verify the failure
modes" (still more of an e2e test than unit test.

PR #120, opened 2025-07-10
Fix HEAD, fix uniqueness checks for typedefs

thockin:
When we have a typedef to a list, and we define list metadata (e.g. listType=map) on the typedef, we were not considering that when we emit uniqueness checks.
    
Additionally, we did not check that the user didn't specify incompatible metadata on both the type and the field.
    
Now we disallow those tags to be in both places, and we look at type-attached metadata if there is not field-attached metadata.  We should never have both.

This uncovered a number of smaller issues which are fixed in different commits in this PR

### DIFF
--- a/staging/src/k8s.io/code-generator/cmd/validation-gen/validation.go
+++ b/staging/src/k8s.io/code-generator/cmd/validation-gen/validation.go
@@ -584,11 +584,11 @@ func (td *typeDiscoverer) discoverStruct(thisNode *typeNode, fldPath *field.Path
 			Path:   childPath,
 		}
 
-		Tags, err := td.validator.ExtractTags(context, memb.CommentLines)
+		tags, err := td.validator.ExtractTags(context, memb.CommentLines)
 		if err != nil {
 			return fmt.Errorf("field %s: %w", childPath.String(), err)
 		}
-		if validations, err := td.validator.ExtractValidations(context, Tags...); err != nil {
+		if validations, err := td.validator.ExtractValidations(context, tags...); err != nil {
 			return fmt.Errorf("field %s: %w", childPath.String(), err)
 		} else if validations.Empty() {
 			klog.V(6).InfoS("no field-attached validations", "field", childPath)


staging/src/k8s.io/code-generator/cmd/validation-gen/validators/each.go
```diff
@@ -254,8 +241,33 @@ func (listValidator) Name() string {
...
+	}
+	// NOTE: We don't really support list-of-list or map-of-list, so this does
+	// not consider the case of ScopeListVal or ScopeMapVal. If we want to
+	// support those, we need to look at this and make sure the paths work the
+	// way we need.
+	if context.Scope == ScopeField {
+		tm := lv.byPath[context.Type.String()]
+		if lm != nil && tm != nil {
+			return Validations{}, fmt.Errorf("found list metadata for both a field and its type: %s", context.Path)
+		}
```
yongruilin, line 261: We do need the test to check the error from GetValidations(). (#113)
So when listTypeValidator process typedef, the `context.Path.String() == context.Type.String()`  right?

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/validators.go
```diff
@@ -97,11 +97,12 @@ type TypeValidator interface {
...
 // RegisterFieldValidator.
 //
-// FieldValidators are always processed after TagValidators, which means that
-// they can "finish" work with data that was collected by TagValidators.
-// FieldValidators are evaluated after all TagValidators and after the field has
-// been fully processed (including all child fields). FieldValidators MUST NOT
-// depend on other FieldValidators having been run already.
+// FieldValidators are always processed after TagValidators and TypeValidators,
+// which means that they can "finish" work with data that was collected by
+// TagValidators. FieldValidators are evaluated after all TagValidators and
```
yongruilin, line 102:
```suggestion
// TagValidators and TypeValidators. FieldValidators are evaluated after all TagValidators and
```

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/each.go
```diff
@@ -44,13 +44,13 @@ func init() {
 	// Lists with list-map semantics are comprised of multiple tags, which need
 	// to share information between them.
 	shared := map[string]*listMetadata{} // keyed by the fieldpath
```
aaron-prindle, line 46: nit: keyed by fieldpath is no longer correct IIUC. Perhaps:
```go
// keyed by path (either field path or type path)
```

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/each.go
```diff
@@ -265,40 +277,60 @@ func (lv listValidator) GetValidations(context Context) (Validations, error) {
...
+		matchArg := lm.makeListMapMatchFunc(nt.Elem)
 		f := Function("listValidator", DefaultFlags, validateUnique, matchArg)
-		result.Functions = append(result.Functions, f)
+		result.AddFunction(f)
 	}
 
 	return result, nil
 }
 
+// make sure a give listMetadata makes sense.
```
aaron-prindle, line 310: nit: given

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/each.go
thockin: ACK

yongruilin (commented):
LGTM

aaron-prindle:
LGTM

thockin:
Quick followup coming in a few mins

thockin:
Pushed with comments fixed

PR #124, opened 2025-07-11
Pass equiv func to SliceItem, small cleanups

thockin:
Rather than send EVEN MORE comments on non-critical things, I just accumulated them.

1) Pass `equiv` to SliceItem(), same as EachVal()
2) minor renames, comments, error strings


aaron-prindle:
LGTM

PR #125, opened 2025-07-12
Add comments to generated code

thockin:
This makes the output longer, but I think it might have helped catch recent bugs that SHOULD have been caught by inspection.

Not urgent, but touches a lot of places.


jpbetz (approved):
This all makes sense to me.  +100 to the `WithComment` function. That's really handy.

PR #126, opened 2025-07-13
Cleanup: Small lint things

thockin:
1) rename receiver var s -> b for ValidationTestBuilder

2) be explicit about the testing.T, rather than embed it


thockin:
We should leave this until after code-freeze.

PR #132, opened 2025-07-17
Refactor Union and Item validation after #128

thockin:
Followup to #128 resolving the issues I flagged (and more).  This could be broken up, but each commit stands alone, more or less.

It took a few tries to find a model that wasn't overly baroque, but I think this hits a decent balance.  Very open to changes in the approach.

@aaron-prindle

### DIFF
--- a/staging/src/k8s.io/code-generator/cmd/validation-gen/validators/item.go
+++ b/staging/src/k8s.io/code-generator/cmd/validation-gen/validators/item.go
@@ -261,8 +261,8 @@ func (iv itemValidator) GetValidations(context Context) (Validations, error) {
 			Scope:      ScopeListVal,
 			Type:       elemT,
 			Path:       context.Path.Key(subContextPath),
+			Member:     nil, // NA for list items
 			ParentPath: context.Path,
-			Member:     nil,
 		}
 
 		validations, err := iv.validator.ExtractValidations(subContext, item.valueTag)
--- a/staging/src/k8s.io/code-generator/cmd/validation-gen/validators/item.go
+++ b/staging/src/k8s.io/code-generator/cmd/validation-gen/validators/item.go
@@ -187,7 +187,7 @@ func (iv *itemValidator) Init(cfg Config) {
 }
 
 func (itemValidator) Name() string {
-	return "itemFieldValidator"
+	return "itemValidator"
 }
 
 var (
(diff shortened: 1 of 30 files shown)


thockin:
Rebased

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/union.go
```diff
@@ -178,28 +172,34 @@ func (umtv unionMemberTagValidator) Docs() TagDoc {
...
 
-	// itemMatchers stores matcher criteria for list item unions.
-	// key is the virtual field path (eg: "<path>/Pipeline.Tasks[{"name": "succeeded"}]"),
-	// value is the parsed matcher map (eg: {"name": "succeeded"}).
-	// Represents union members that are list items matching specific criteria
-	itemMatchers map[string]map[string]any
+	// itemMembers stores selection criteria for list-item unions.
+	// The map key is the "field name" (eg: `field[{"name": "succeeded"}]`),
+	// and the value is a list of selection criteria.
+	itemMembers map[string][]ListSelectorTerm
```
aaron-prindle, line 197: nit: might make sense to add in the comments here that itemMembers and fieldMembers are mutually exclusive (code expects them not to both be set)

aaron-prindle:
LGTM to the overall approach, had two small nits

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/item.go
```diff
@@ -68,68 +60,162 @@ func (itemTagValidator) ValidScopes() sets.Set[Scope] {
...
+		f.Cohort = itemKey
+		result.AddFunction(f)
+	}
+
+	return result, nil
+}
+
+func criteriaFromArgs(args []codetags.Arg) ([]keyValuePair, error) {
+	criteria := []keyValuePair{}
+	keysSeen := sets.NewString()
```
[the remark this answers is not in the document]
thockin: oops!  Autocomplete fail.

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/union.go
thockin: Good idea.

PR #137, opened 2025-07-19
Emit ratchet check before, not in, type funcs

### LINKED ISSUE #133: Double-ratche checking for typedefs-to-lists which are listmaps
E.g. staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/tags/item/typedef

```
// +k8s:listType=map    
// +k8s:listMapKey=key    
// +k8s:item(key: "immutable")=+k8s:immutable    
// +k8s:item(key: "validated")=+k8s:validateFalse="item ItemList[key=validated]"    
type ItemList []Item    
```

When there are any field validations (including the list-map must be unique) then the generated code looks like:

```
    // field Struct.TypedefItems
    errs = append(errs,
        func(fldPath *field.Path, obj, oldObj ItemList) (errs field.ErrorList) {
            // don't revalidate unchanged data
            if op.Type == operation.Update && equality.Semantic.DeepEqual(obj, oldObj) {
                return nil
            }
            // call field-attached validations
            // listType=map requires unique keys
            errs = append(errs, validate.Unique(ctx, op, fldPath, obj, oldObj, func(a Item, b Item) bool { return a.Key == b.Key })...)
            // call the type's validation function
            errs = append(errs, Validate_ItemList(ctx, op, fldPath, obj, oldObj)...)
            return                                                                                                                                                       
        }(fldPath.Child("typedefItems"), obj.TypedefItems, safe.Field(oldObj, func(oldObj *Struct) ItemList { return oldObj.TypedefItems }))...)
```

The problem is that `Validate_ItemList()` will *ALSO* call `DeepEqual`.  From a completeness POV it needs to, since it COULD be called from anywhere and should not rely on the fact that it was already checked for uniqueness.  But in examples like this it gets called twice. :(

Perhaps we should do something like:

```
// Validate_ItemList validates an instance of ItemList according
// to declarative validation rules in the API schema.
func Validate_ItemList(ctx context.Context, op operation.Operation, fldPath *field.Path, obj, oldObj ItemList) (errs field.ErrorList) {
    // don't revalidate unchanged data
    if op.Type == operation.Update && equality.Semantic.DeepEqual(obj, oldObj) {
        return nil
    }
    return validate_internal_ItemList(...)
}

func validate_internal_ItemList(ctx context.Context, op operation.Operation, fldPath *field.Path, obj, oldObj ItemList) (errs field.ErrorList) {
    errs = append(errs, validate.SliceItem(ctx, op, fldPath, obj, oldObj, func(item *Item) bool { return item.Key == "immutable" }, validate.DirectEqual, validate.ImmutableByCompare)...)
    errs = append(errs, validate.SliceItem(ctx, op, fldPath, obj, oldObj, func(item *Item) bool { return item.Key == "validated" }, validate.DirectEqual, func(ctx context.Context, op operation.Operation, fldPath *field.Path, obj, oldObj *Item) field.ErrorList {
        return validate.FixedResult(ctx, op, fldPath, obj, oldObj, false, "item ItemList[key=validated]")
    })...)
    // listType=map requires unique keys
    errs = append(errs, validate.Unique(ctx, op, fldPath, obj, oldObj, func(a Item, b Item) bool { return a.Key == b.Key })...)

    return errs                                                                                                                                                                                                     
}
```

External callers call the public function, and internals call the internal?  But that doesn't work for cross-pkg calls, so maybe they are both public symbols?  Or maybe this just doesn't matter.

### LINKED ISSUE #136: Not doing ratchet-check on root fields
@yongruilin 

Validation functions look like:

```
func Validate_Foo(ctx context.Context, op operation.Operation, fldPath *field.Path, obj, oldObj *corev1.Foo) (errs field.ErrorList) {                                           
      errs = append(errs,
        func(fldPath *field.Path, obj, oldObj *corev1.FooSpec) (errs field.ErrorList) {
            errs = append(errs, Validate_FooSpec(ctx, op, fldPath, obj, oldObj)...)
            return
        }(fldPath.Child("spec"), &obj.Spec, safe.Field(oldObj, func(oldObj *corev1.Foo) *corev1.FooSpec { return &oldObj.Spec }))...)
}

func Validate_FooSpec(ctx context.Context, op operation.Operation, fldPath *field.Path, obj, oldObj *corev1.FooSpec) (errs field.ErrorList) {
    // field corev1.FooSpec.Field
    errs = append(errs,
        func(fldPath *field.Path, obj, oldObj *int32) (errs field.ErrorList) {
            // don't revalidate unchanged data
            if op.Type == operation.Update && (obj == oldObj || (obj != nil && oldObj != nil && *obj == *oldObj)) {
                return nil
            }
            // call field-attached validations
            return
        }(fldPath.Child("spec"), obj.Field, safe.Field(oldObj, func(oldObj *corev1.FooSpec) *int32 { return oldObj.Field }))...)
}
```

Note that it is not doing a ratchet-check on `spec`.  I think this is because spec has no field-attached validations.

thockin:
In particular, we were not emitting the check for fields which have a type func and no field-attached validation.  For lists we were doing both.  This could go inside the type func but then we end up with a double ratchet sometimes.  See #133 

This proposes that the protocol for ratchet checking is that the caller does a ratchet-check before calling a type's validation function.

Fixes #136
Fixes #133

### DIFF
--- a/staging/src/k8s.io/code-generator/cmd/validation-gen/validation.go
+++ b/staging/src/k8s.io/code-generator/cmd/validation-gen/validation.go
@@ -955,9 +955,7 @@ func (g *genValidations) emitValidationForChild(c *generator.Context, thisChild 
 		default:
 			panic(fmt.Sprintf("unexpected type-validations on type %v, kind %s", thisNode.valueType, thisNode.valueType.Kind))
 		}
-		sw.Do("// type $.inType|raw$\n", targs)
 		emitComments(validations.Comments, sw)
-		emitRatchetingCheck(c, thisNode.valueType, sw)
 		emitCallsToValidators(c, validations.Functions, sw)
 		if thisNode.valueType.Kind == types.Alias {
 			underlyingNode := thisNode.underlying.node
@@ -1008,6 +1006,49 @@ func (g *genValidations) emitValidationForChild(c *generator.Context, thisChild 
 			// Accumulate into a buffer so we don't emit empty functions.
 			buf := bytes.NewBuffer(nil)
 			bufsw := sw.Dup(buf)
+
+			// On ratcheting checks:
+			//
+			// We emit ratcheting checks ONLY for struct fields and ONLY when
+			// that field has some validations to call.
+			//
+			// We DO NOT emit ratchet checks inside type-specific validation
+			// functions, because that leads to repeated ratchet checking which
+			// is almost never useful work (keep reading).
+			//
+			// The consequence of this is that a caller of a type's validation
+			// function is assumed to have already done a ratchet check (which
+			// is true for all generated code (except root types, keep
+			// reading)). For struct types (our most common case), the type's
+			// function will do ratchet checks on each sub-field anyway.
+			//
+			// This leaves one case where validation is executed unilaterally:
+			// non-pre-checked calls of validation functions for types which have
+			// type-attached validations.  This can happen in two cases:
+			//   1. an external caller of a type's validation function
+			//   2. the generated register function for a package which calls a
+			//      root-type's validation function
+			//
+			// TODO: We are leaving this as a problem for the future. If we
+			// find that we have a root type which has type-attached validation
+			// AND that validation is being ratcheted, then we will need to
+			// address this. Some options:
+			//   1. emit a ratchet check in the package's register function
+			//   2. emit a ratchet check in the type's validation function
+			//      (IFF it has type-attached validation, perhaps only for root
+			//      types)
+			//   3. emit both "safe" (ratchet check the whole object) and
+			//      "fast" (assume the object was already ratchet checked)
+			//      forms of each type's validation function, so that the
+			//      generated code can call the "fast" form while external code
+			//      calls the "safe" form.
+			//   4. implement depth-first traversal of validation, where each
+			//      function returns an additional bool indicating "something
+			//      changed", which gets propagated up the caller to decide if
+			//      it needs to do higher-level validations (e.g. if any field
+			//      in a struct changes, the struct's type-attached validations
+			//      need to be executed, but if no fields changed they can be
+			//      skipped).
 
 			validations := fld.fieldValidations
 			fldRatchetingChecked := false
--- a/staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/all_types_match/with_type_validations/zz_generated.validations.go
+++ b/staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/all_types_match/with_type_validations/zz_generated.validations.go
@@ -59,11 +59,6 @@ func RegisterValidations(scheme *testscheme.Scheme) error {
 // Validate_E1 validates an instance of E1 according
 // to declarative validation rules in the API schema.
 func Validate_E1(ctx context.Context, op operation.Operation, fldPath *field.Path, obj, oldObj *E1) (errs field.ErrorList) {
-	// type E1
-	// don't revalidate unchanged data
-	if op.Type == operation.Update && (obj == oldObj || (obj != nil && oldObj != nil && *obj == *oldObj)) {
-		return nil
-	}
 	errs = append(errs, validate.FixedResult(ctx, op, fldPath, obj, oldObj, false, "type E1")...)
 
 	return errs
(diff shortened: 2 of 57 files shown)


thockin:
Let's get Aaron's changes in first!

staging/src/k8s.io/code-generator/cmd/validation-gen/validation.go
```diff
@@ -1014,9 +1012,9 @@ func (g *genValidations) emitValidationForChild(c *generator.Context, thisChild
 			if !validations.Empty() {
 				emitComments(validations.Comments, bufsw)
 				emitRatchetingCheck(c, fld.childType, bufsw)
+				fldRatchetingChecked = true
```
yongruilin, line 1015: Would this move impact the generated code?

staging/src/k8s.io/code-generator/cmd/validation-gen/validation.go
```diff
@@ -955,9 +955,7 @@ func (g *genValidations) emitValidationForChild(c *generator.Context, thisChild
 		default:
 			panic(fmt.Sprintf("unexpected type-validations on type %v, kind %s", thisNode.valueType, thisNode.valueType.Kind))
 		}
-		sw.Do("// type $.inType|raw$\n", targs)
 		emitComments(validations.Comments, sw)
-		emitRatchetingCheck(c, thisNode.valueType, sw)
```
yongruilin, line 960: So, no ratcheting on typeValidation?
Do we need to consider the case that root type is embedded in other type?

```
// +k8s:validateFalse="type DirectComparableStruct"
type DirectComparableStruct struct {
	TypeMeta int
	// +k8s:validateFalse="field intField"
	IntField int `json:"intField"`
}

type StructStruct struct {
	TypeMeta int
	// +k8s:validateFalse="field directComparableStructField"
	DirectComparableStructField DirectComparableStruct `json:"directComparableStructField"`
}
```

thockin: I think that this case would have `StructStruct` do the ratchet check on `DirectComparableStructField` before calling `Validate_DirectComparableStruct()`.

IOW, the calling conventions is: If you call a validation function it will ratchet-check individual fields but will not ratchet-check it's whole self.

yongruilin: i means typeValidation's ratcheting check always execute by the caller, and we assume that there should not be a case that root type itself has a type validation attached. Am I understanding it correctly?

yongruilin: > Even if a root type has a ratchet check, the per-field ratcheting will handle it.

I think it removes the ratcheting on the root type's `typeValidation`. e.g. we attached a k8s:validateFalse directly to the root type. Then it would just generate
```
func Validate_RootStruct(ctx context.Context, op operation.Operation, fldPath *field.Path, obj, oldObj *DirectComparableStruct) (errs field.ErrorList) {
	errs = append(errs, validate.FixedResult(ctx, op, fldPath, obj, oldObj, false, "type RootStruct")...)
	errs = append(errs,
		func(fldPath *field.Path, obj, oldObj *int) (errs field.ErrorList) {
     ...
     ...
```

But I think k8s:validateFalse or any other type validation attached on a root type only happen in testing.

This PR LGTM to remove the double-ratcheting.

thockin: > I think it removes the ratcheting on the root type's typeValidation

You are right.  If someone is calling Validate_Pod(), the pod is going to be validated.  It will be ratchet-checked, field-by-field, but the code which runs over those fields WILL be called.  This goes back to the discussion about how to implement ratcheting efficiently.  In a perfect design, every field would be ratchet-checked exactly once.  We trade-off simplicity in implementation for performance, so that isn't "perfect".

The question is whether it's better to have it be obvious (this PR) or occasionally redundant (prior state).

thockin:
Rebased

thockin:
I went to write the comment:

```
            // On ratcheting: We emit ratcheting checks a) ony for struct
            // fields; and b) only when that field has some validations to
            // call. We do not emit ratchet checks inside type-specific
            // validation functions, because that leads to repeated ratchet
            // checking. The consequence of this is that a caller of a type's
            // validation function is assumed to have already done a ratchet
            // check (which is true for all generated code except root types).
            // For struct types (our most common case), the type's function
            // will do ratchet checks on each sub-field anyway.
            //
            // This leaves one case where validation is executed unilaterally:
            // non-pre-checked calls of validation functions for types which have
            // type-attached validations.  This can happen in two cases:
            // 1. Some external caller of a type's validation function
            // 2. The generated resgister function for a package which calls a
            //    root-type's validation function.
```

And it made me wonder if we should handle the 2nd case better?.  For it to matter we would need:
1) a root type with type-attached validation; AND
2) one of those type-attached validations is being ratcheted

Depending on how we do cross-field things, I could see (1) being true but (2) is less likely.

We could change the Register function to always do a ratchet-check of root types?

We could change type-functions to know "is a root type" and do a ratchet check IFF there are type-attached validations.

We could just document it and see if it matters.

I think the last one is right.

yongruilin (approved):
LGTM. I will merge it once the comment is in.

PR #139, opened 2025-07-21
Rename tests before they get copied more widely

thockin:


pkg/registry/core/replicationcontroller/declarative_validation_test.go
```diff
@@ -31,7 +31,7 @@ import (
 	"k8s.io/utils/ptr"
 )
 
-func TestDeclarativeValidateForDeclarative(t *testing.T) {
+func TestDeclarativeValidate(t *testing.T) {
```
aaron-prindle, line 34: nit: This is "TestDeclarativeValidate" but for `pkg/registry/certificates/certificates/declarative_validation_test.go` a similar test is:
TestDeclarativeValidation

Might be missing a difference in the tests but could make sense to align the naming

aaron-prindle:
LGTM, merging

PR #142, opened 2025-07-26
Finish "IP sloppy" removal

thockin:
In a previous PR we renamed format "k8s-ip-sloppy" to "k8s-ip", but we did not fix the actual validation.  This set of commits changes the validation to follow the strict rules in the handwritten validation.  We can't convert it all at the same time because there is a gate, so this adds some comments.


thockin:
Rebased

Issue #145, opened 2025-07-30
Add Stability Level To Tag Validators (Alpha, Beta, GA) so users can know which tags are stable vs experimental


aaron-prindle:
Issue tracks adding adding a Stability Level To Tag Validators (Alpha, Beta, GA)  Docs() method so users can know which tags are stable vs experimental

PR #146, opened 2025-07-31
Add stability level in the tags doc.

lalitc375:
Fixes https://github.com/jpbetz/validation-gen/issues/145

This PR introduces a StabilityLevel to all validation tags to indicate their maturity. This helps users understand the stability of each validation rule and whether its API is subject to change.

A StabilityLevel field has been added to the TagDoc struct and can be either Stable or Alpha

- **Stable**: The tag's API and behavior are considered stable and are not expected to change.
- **Alpha**: The tag is considered experimental, and its API or behavior may change in the future.

This change systematically updates all existing validators to set an appropriate stability level, which will be surfaced in the generated documentation. This provides clear guidance to developers about which validation rules are safe for long-term use.

### DIFF
--- a/staging/src/k8s.io/code-generator/cmd/validation-gen/validators/each.go
+++ b/staging/src/k8s.io/code-generator/cmd/validation-gen/validators/each.go
@@ -167,9 +167,10 @@ func (lttv listTypeTagValidator) GetValidations(context Context, tag codetags.Ta
 
 func (lttv listTypeTagValidator) Docs() TagDoc {
 	doc := TagDoc{
-		Tag:         lttv.TagName(),
-		Scopes:      lttv.ValidScopes().UnsortedList(),
-		Description: "Declares a list field's semantic type.",
+		Tag:            lttv.TagName(),
+		StabilityLevel: Stable,
+		Scopes:         lttv.ValidScopes().UnsortedList(),
+		Description:    "Declares a list field's semantic type.",
 		Payloads: []TagPayloadDoc{{
 			Description: "<type>",
 			Docs:        "atomic | map | set",
@@ -228,9 +229,10 @@ func (lmktv listMapKeyTagValidator) GetValidations(context Context, tag codetags
 
 func (lmktv listMapKeyTagValidator) Docs() TagDoc {
 	doc := TagDoc{
-		Tag:         lmktv.TagName(),
-		Scopes:      lmktv.ValidScopes().UnsortedList(),
-		Description: "Declares a named sub-field of a list's value-type to be part of the list-map key.",
+		Tag:            lmktv.TagName(),
+		StabilityLevel: Stable,
+		Scopes:         lmktv.ValidScopes().UnsortedList(),
+		Description:    "Declares a named sub-field of a list's value-type to be part of the list-map key.",
 		Payloads: []TagPayloadDoc{{
 			Description: "<field-json-name>",
 			Docs:        "The name of the field.",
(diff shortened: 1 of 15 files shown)


aaron-prindle:
@lalitc375 Currently this PR has ~117 commits, I think you might need to rebase this

lalitc375:
> @lalitc375 Currently this PR has ~117 commits, I think you might need to rebase given the recent validation-gen rebase that was done onto kubernetes/master. I believe this is the only relevant commit for the PR:
> https://github.com/jpbetz/validation-gen/commit/b0735f9ae1a537a76f44a2e64b3b6b1b98020733

My bad, Somehow I missed it.

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/format.go
```diff
@@ -93,9 +93,10 @@ func getFormatValidationFunction(format string) (FunctionGen, error) {
 
 func (ftv formatTagValidator) Docs() TagDoc {
 	return TagDoc{
-		Tag:         ftv.TagName(),
-		Scopes:      ftv.ValidScopes().UnsortedList(),
-		Description: "Indicates that a string field has a particular format.",
+		Tag:            ftv.TagName(),
+		Scopes:         ftv.ValidScopes().UnsortedList(),
+		StabilityLevel: Stable,
```
aaron-prindle, line 98: For +k8s:format to be `Stable` I would believe that would mean any usage of the tag is stable.  Do we believe this will be the case for the various format tags we plan to support?  Do we need any kind of metered way of ensuring format tags once upstreamed are not modified or none are removed given this?

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/each.go
```diff
@@ -180,9 +180,10 @@ func (lttv listTypeTagValidator) GetValidations(context Context, tag codetags.Ta
 
 func (lttv listTypeTagValidator) Docs() TagDoc {
 	doc := TagDoc{
-		Tag:         lttv.TagName(),
-		Scopes:      lttv.ValidScopes().UnsortedList(),
-		Description: "Declares a list field's semantic type.",
+		Tag:            lttv.TagName(),
+		StabilityLevel: Stable,
```
aaron-prindle, line 184: nit: perhaps not necessary for this PR but it might make sense to at tag registration time enforce that a stability level is set on a tag being registered and the value is an allowed value if we plan to use Stability level on each tag for documentation generation, etc.

lalitc375:
> LGTM for the PR here adding the `StabilityLevel` field and the levels agreed upon in the "DV Tags Stability Levels" document. If not already created can we add tracking issues for:
> 
> * enforcing stability level on all tags
> * documenting the stability level and what the guarantees are for Stable and Alpha

I have created the tracking issues you requested . The new issues are #161 for enforcing the stability level on all tags, and #162 for documenting the stability level guarantees.

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/format.go
lalitc375: > As all formats are stable now I believe this can be "stable" but we would need a method to update this likely per acceptable payload argument IIUC from the above comments.

Yes, We will have to revisit this while adding new formats , then we add stability levels for payloads.

aaron-prindle:
LGTM, merging

PR #150, opened 2025-08-08
Use tag chaining for options

jpbetz:
This came up in discussion: https://github.com/jpbetz/validation-gen/discussions/149#discussioncomment-14010598

I don't remember why we didn't use chaining originally. Did we not have the chaining built out then?  Whatever the reason, this switches to chaining.

### DIFF
--- a/staging/src/k8s.io/apimachinery/pkg/api/validate/options.go
+++ b/staging/src/k8s.io/apimachinery/pkg/api/validate/options.go
@@ -25,14 +25,8 @@ import (
 
 // IfOption conditionally evaluates a validation function. If the option and enabled are both true the validator
 // is called. If the option and enabled are both false the validator is called. Otherwise, the validator is not called.
-func IfOption[T any](
-	ctx context.Context,
-	op operation.Operation,
-	fldPath *field.Path,
-	value, oldValue *T,
-	optionName string,
-	enabled bool,
-	validator func(context.Context, operation.Operation, *field.Path, *T, *T) field.ErrorList,
+func IfOption[T any](ctx context.Context, op operation.Operation, fldPath *field.Path, value, oldValue *T,
+	optionName string, enabled bool, validator func(context.Context, operation.Operation, *field.Path, *T, *T) field.ErrorList,
 ) field.ErrorList {
 	if op.HasOption(optionName) == enabled {
 		return validator(ctx, op, fldPath, value, oldValue)
--- a/staging/src/k8s.io/code-generator/cmd/validation-gen/validators/options.go
+++ b/staging/src/k8s.io/code-generator/cmd/validation-gen/validators/options.go
@@ -55,7 +55,7 @@ func (ifOptionTagValidator) ValidScopes() sets.Set[Scope] {
 }
 
 var (
-	ifOptionSubfield = types.Name{Package: libValidationPkg, Name: "IfOption"}
+	ifOption = types.Name{Package: libValidationPkg, Name: "IfOption"}
 )
 
 func (iotv ifOptionTagValidator) GetValidations(context Context, tag codetags.Tag) (Validations, error) {
@@ -64,14 +64,8 @@ func (iotv ifOptionTagValidator) GetValidations(context Context, tag codetags.Ta
 	if validations, err := iotv.validator.ExtractValidations(context, *tag.ValueTag); err != nil {
 		return Validations{}, err
 	} else {
-		var tagName string
-		if iotv.enabled {
-			tagName = ifOptionEnabledTag
-		} else {
-			tagName = ifOptionDisabledTag
-		}
 		for _, fn := range validations.Functions {
-			f := Function(tagName, fn.Flags, ifOptionSubfield, optionName, iotv.enabled, WrapperFunction{Function: fn, ObjType: context.Type})
+			f := Function(iotv.TagName(), fn.Flags, ifOption, optionName, iotv.enabled, WrapperFunction{Function: fn, ObjType: context.Type})
 			result.Variables = append(result.Variables, validations.Variables...)
 			result.AddFunction(f)
 		}


staging/src/k8s.io/code-generator/cmd/validation-gen/validators/options.go
```diff
@@ -53,19 +54,26 @@ func (ifOptionTagValidator) ValidScopes() sets.Set[Scope] {
...
+
 func (iotv ifOptionTagValidator) GetValidations(context Context, tag codetags.Tag) (Validations, error) {
 	optionName := tag.Args[0].Value
 	result := Validations{}
 	if validations, err := iotv.validator.ExtractValidations(context, *tag.ValueTag); err != nil {
 		return Validations{}, err
 	} else {
+		var tagName string
+		if iotv.enabled {
+			tagName = ifOptionEnabledTag
```
thockin, line 69: `tagName := iotv.TagName()`

staging/src/k8s.io/apimachinery/pkg/api/validate/options.go
```diff
@@ -0,0 +1,41 @@
...
+	"context"
+
+	"k8s.io/apimachinery/pkg/api/operation"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+)
+
+// IfOption conditionally evaluates a validation function. If the option and enabled are both true the validator
+// is called. If the option and enabled are both false the validator is called. Otherwise, the validator is not called.
+func IfOption[T any](
+	ctx context.Context,
```
thockin, line 29: Stylistic - most of the validation funcs seem to keep ctx, op, fldPath, value, oldValue on one line (standard args) and some wrap to a newline for extended args -- do we care about any sort of convention here?

thockin (commented):
This LGTM.  Will leave open for a bit for Aaron and Yongrui to have a look.

Applying this to the name-ratcheting idea:

```
diff --git pkg/apis/core/v1/zz_generated.validations.go pkg/apis/core/v1/zz_generated.validations.go
index eaeab7f8dc3..7e1c30f8c57 100644
--- pkg/apis/core/v1/zz_generated.validations.go
+++ pkg/apis/core/v1/zz_generated.validations.go
@@ -76,7 +76,12 @@ func Validate_ReplicationController(ctx context.Context, op operation.Operation,
 				if e := validate.Subfield(ctx, op, fldPath, obj, oldObj, "name", func(o *metav1.ObjectMeta) *string { return &o.Name }, validate.DirectEqualPtr, validate.OptionalValue); len(e) != 0 {
 					return // do not proceed
 				}
-				errs = append(errs, validate.Subfield(ctx, op, fldPath, obj, oldObj, "name", func(o *metav1.ObjectMeta) *string { return &o.Name }, validate.DirectEqualPtr, validate.LongName)...)
+				errs = append(errs, validate.Subfield(ctx, op, fldPath, obj, oldObj, "name", func(o *metav1.ObjectMeta) *string { return &o.Name }, validate.DirectEqualPtr, func(ctx context.Context, op operation.Operation, fldPath *field.Path, obj, oldObj *string) field.ErrorList {
+					return validate.IfOption(ctx, op, fldPath, obj, oldObj, "RelaxedServiceNameValidation", false, validate.ShortName)
+				})...)
+				errs = append(errs, validate.Subfield(ctx, op, fldPath, obj, oldObj, "name", func(o *metav1.ObjectMeta) *string { return &o.Name }, validate.DirectEqualPtr, func(ctx context.Context, op operation.Operation, fldPath *field.Path, obj, oldObj *string) field.ErrorList {
+					return validate.IfOption(ctx, op, fldPath, obj, oldObj, "RelaxedServiceNameValidation", true, validate.LongName)
+				})...)
 			}()
 			return
 		}(fldPath.Child("metadata"), &obj.ObjectMeta, safe.Field(oldObj, func(oldObj *corev1.ReplicationController) *metav1.ObjectMeta { return &oldObj.ObjectMeta }))...)
diff --git staging/src/k8s.io/api/core/v1/types.go staging/src/k8s.io/api/core/v1/types.go
index 091f23af143..ed011a4d4c4 100644
--- staging/src/k8s.io/api/core/v1/types.go
+++ staging/src/k8s.io/api/core/v1/types.go
@@ -5579,7 +5579,8 @@ type ReplicationController struct {
 	// Standard object's metadata. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#metadata
 	// +optional
 	// +k8s:subfield(name)=+k8s:optional
-	// +k8s:subfield(name)=+k8s:format=k8s-long-name
+	// +k8s:subfield(name)=+k8s:ifOptionDisabled(RelaxedServiceNameValidation)=+k8s:format=k8s-short-name
+	// +k8s:subfield(name)=+k8s:ifOptionEnabled(RelaxedServiceNameValidation)=+k8s:format=k8s-long-name
 	metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
 
 	// Spec defines the specification of the desired behavior of the replication controller.
```

jpbetz:
Added a subfield test as well.

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/options.go
```diff
@@ -53,19 +54,20 @@ func (ifOptionTagValidator) ValidScopes() sets.Set[Scope] {
 	return ifOptionTagValidScopes
 }
 
+var (
+	ifOptionSubfield = types.Name{Package: libValidationPkg, Name: "IfOption"}
```
aaron-prindle, line 58: nit: might make sense to have a different name here as this is used for any tag that could be chained of the conditional IIUC not just subfield.

staging/src/k8s.io/apimachinery/pkg/api/validate/options_test.go
```diff
@@ -0,0 +1,99 @@
+/*
+Copyright 2024 The Kubernetes Authors.
```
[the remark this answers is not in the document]
thockin: In general we do not want to bump copyright dates.  They represent when it was first produced

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/options.go
thockin: This is specifically checking the options field, isn't it?

yongruilin (commented):
This PR LGTM. I have 2 things in my mind:

1. I assume the usage would be have in call site, having featuregate check and hardcoded the "option" string, e.g.
```
		var dvOptions []string
		if utilfeature.DefaultFeatureGate.Enabled(features.FeatureX) {
			dvOptions = []string{"FeatureX"}
		}
		// Run declarative validation with panic recovery
		declarativeErrs := rest.ValidateDeclaratively(ctx, legacyscheme.Scheme, controller, rest.WithTakeover(takeover), rest.WithOptions(dvOptions))
```
But do we have use case other than featuregate? Is it possible to have the featuregate check logic inside `validate.IfOption[T]`, so that we don't need to?

2. Currently we haven't migrated any real-world case, once we have, do we need to update our testing pattern to add another dimension with featuregate enable/disable?

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/options.go
aaron-prindle: This is renamed now s/ifOptionSubfield/ifOption.  Marking resolved

aaron-prindle:
LGTM, merging

Issue #153, opened 2025-08-11
"Not found" chained tags are not handled safely


jpbetz:
xref: https://github.com/jpbetz/validation-gen/pull/152#issuecomment-3175722095

We should improve our chained tag processing to consistently report a clear error and not panic for such cases.

PR #158, opened 2025-08-12
Fix validation-gen panic on unknown tags during code generation

yongruilin:
Fix: #153 

I originally added the nil check to for-loop in ExtractValidations here
```go
// Run tag-validators first.
phases := reg.sortTagsIntoPhases(tags)
for _, tags := range phases {
		for _, tag := range tags {
			tv := reg.tagValidators[tag.Name]
-------> // nil check here
```
when I found out that we should check the tags in `sortTagsIntoPhases`. Then I finally moved all the checks upfront and aggregated any "not found" errors.

### DIFF
--- a/staging/src/k8s.io/code-generator/cmd/validation-gen/validators/registry.go
+++ b/staging/src/k8s.io/code-generator/cmd/validation-gen/validators/registry.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"slices"
 	"sort"
+	"strings"
 	"sync"
 	"sync/atomic"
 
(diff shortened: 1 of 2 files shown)


staging/src/k8s.io/code-generator/cmd/validation-gen/validators/registry.go
```diff
@@ -147,11 +148,27 @@ func (reg *registry) ExtractValidations(context Context, tags ...codetags.Tag) (
 	}
 	validations := Validations{}
 
+	// Check all tags first for tag processing issues
+	var errors []string
+	for _, tag := range tags {
```
jpbetz, line 153: Also check the tag.ValueTag's (recursively)?

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/registry_test.go
```diff
@@ -0,0 +1,115 @@
...
+)
+
+func TestExtractValidations_TagValidator(t *testing.T) {
+	tests := []struct {
+		name    string
+		tags    []codetags.Tag
+		setup   func(*registry)
+		wantErr string
+	}{
+		{
```
jpbetz, line 35: + chained tag test cases

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/registry.go
yongruilin: Updated.

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/registry_test.go
yongruilin: Added the testcase.

jpbetz:
LGTM, waiting for green CI to merge

PR #169, opened 2025-08-20
feat(validation): Add k8s-label-value format validation

lalitc375:
feat(validation): Add k8s-label-value format validation

This PR introduces a new format validator for Kubernetes label values and adds a comprehensive set of test cases.


staging/src/k8s.io/apimachinery/pkg/api/validate/strfmt.go
```diff
@@ -65,3 +65,23 @@ func LongName[T ~string](_ context.Context, op operation.Operation, fldPath *fie
...
+//
+// All errors returned by this function will be "invalid" type errors. If the
+// caller wants better errors, it must take responsibility for checking things
+// like required/optional and max-length.
+func LabelValue[T ~string](_ context.Context, op operation.Operation, fldPath *field.Path, value, _ *T) field.ErrorList {
+	if value == nil {
+		return nil
+	}
+	var allErrs field.ErrorList
+	for _, msg := range content.IsLabelValue((string)(*value)) {
```
aaron-prindle, line 83: DRA has 4 instances: https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/resource/validation/validation.go (Ctrl-F ValidateLabelName)

of using `ValidateLbelName` but no instances of using content.IsLabelValue.  DRA two instances of this method:

```go

func validateLabelValue(value string, fldPath *field.Path) field.ErrorList {
	var allErrs field.ErrorList

	// There's no metav1validation.ValidateLabelValue.
	for _, msg := range validation.IsValidLabelValue(value) {
		allErrs = append(allErrs, field.Invalid(fldPath, value, msg))
	}

	return allErrs
}
```
https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/resource/validation/validation.go#L1359-L1368

but this does not use content.IsLabelValue.  Can you confirm this tags could be used on DRA types and confirm any deltas.  Is content.IsLabelValue the correct method to use here from the ones listed above?

lalitc375: Thanks for the detailed review. You've raised a great point.

You are correct that DRA currently uses `validation.IsValidLabelValue`. I looked into the source code and found that `validation.IsValidLabelValue` is a deprecated alias for `content.IsLabelValue`.

```go
// staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go

// IsValidLabelValue tests whether the value passed is a valid label value. If
// the value is not valid, a list of error strings is returned. Otherwise an
// empty list (or nil) is returned.
// Deprecated: Use k8s.io/apimachinery/pkg/api/validate/content.IsLabelValue instead.
var IsValidLabelValue = content.IsLabelValue
```

Given this, `content.IsLabelValue` is indeed the correct, non-deprecated function to use, and there are no functional differences ("deltas") between the two.

To answer your other question, the specific fields in DRA I plan to apply the `k8s:format=k8s-label-value` tag to are `DeviceTaint.Value` and `DeviceToleration.Value`.

thockin: As part of this whole DV effort I started a migration.  The goals are:

1) have one library which implements all of our string-format checking which is not using API errors ("content") and has great tests and consistent naming/signatures
2) remove format checking from ALL other util libraries and leave forwarding functions
3) wrap "content" into "validate" functions which produce API errors
4) remove format API errors from ALL other util libraries
5) eventually remove all the other util libraries in favor of "validate" and "content"

I even left README files!

So the right thing to do is to use "content" functions and to migrate in that direction.  Death to all validation/util packages - we have about 8 of them.

It's a side-quest, but I think we can do it while doing the main mission :)

thockin: Also, IsQualifiedName could -> IsLabelKey ?

staging/src/k8s.io/apimachinery/pkg/api/validate/strfmt.go
```diff
@@ -65,3 +65,23 @@ func LongName[T ~string](_ context.Context, op operation.Operation, fldPath *fie
...
 }
+
+// LabelValue verifies that the specified value is a valid label value.
+//   - can be empty
+//   - must be less than 64 characters long
+//   - must start and end with alphanumeric characters
+//   - must contain only alphanumeric characters, dashes, underscores, or dots
+//
+// All errors returned by this function will be "invalid" type errors. If the
+// caller wants better errors, it must take responsibility for checking things
```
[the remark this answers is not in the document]
lalitc375: Other methods in this file, also have the same notice. I think we can remove this notice from here. I will remove it.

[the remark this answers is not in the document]
thockin: Yeah, I think we should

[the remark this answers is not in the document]
lalitc375: Done

staging/src/k8s.io/apimachinery/pkg/api/validate/strfmt.go
```diff
@@ -82,3 +82,19 @@ func LongNameCaseless[T ~string](_ context.Context, op operation.Operation, fldP
 	}
 	return allErrs
 }
+
+// LabelValue verifies that the specified value is a valid label value.
+//   - can be empty
+//   - must be less than 64 characters long
```
thockin, line 88: SUPER picky, but usually we write this as inclusive - must be no more than 63 characters

aaron-prindle:
LGTM, merging

PR #177, opened 2025-08-29
Add validation_identifier for DV mismatch metrics

lalitc375:
Add validation identifier in mismatch metrics.

* Also improve tests for metrics.

### DIFF
--- a/pkg/registry/certificates/certificates/strategy.go
+++ b/pkg/registry/certificates/certificates/strategy.go
@@ -121,12 +121,13 @@ func (csrStrategy) Validate(ctx context.Context, obj runtime.Object) field.Error
 	if utilfeature.DefaultFeatureGate.Enabled(features.DeclarativeValidation) {
 		// Determine if takeover is enabled
 		takeover := utilfeature.DefaultFeatureGate.Enabled(features.DeclarativeValidationTakeover)
+		validation_identifier := "certificate_signing_request_create"
 
 		// Run declarative validation with panic recovery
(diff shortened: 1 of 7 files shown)


richabanker:
Even though this metric is at BETA level which is [allowed to be modified by adding new labels](https://kubernetes.io/docs/reference/instrumentation/metrics/#list-of-beta-kubernetes-metrics), here the metric type is also being changed from Counter to CounterVec - which I think is more of a breaking change since users of the previous metric who were using queries like  `rate(apiserver_validation_declarative_validation_mismatch_total[5m])` will start seeing multiple time series rather than a single one as they would have before. 

Also looking at the [KEP for extending metrics stability](https://github.com/kubernetes/enhancements/tree/master/keps/sig-instrumentation/3498-extending-stability#beta-metrics) which is where we introduced the BETA level, it states that BETA metrics are "forward compatible" with respect to queries and alerts. Changing a metric's type from a Counter to a CounterVec is a more fundamental change than adding a label. It alters the underlying data structure from a single time series to a collection of time series. This change would break existing queries and alerts that are not written to expect a vector, ultimately violating the KEP's forward compatibility guarantee. 

So I think the safest option is to deprecate the older metric and create a new one with the right type
(Side note for me: Update the docs to mention metric types should not change from BETA+ levels.)

lalitc375:
@richabanker Thank you for providing  the feedback. I have created  new metrics to capture validation identifier in [133946](https://github.com/kubernetes/kubernetes/pull/133946) . Let's discuss this further it in the k/k PR.

PR #180, opened 2025-09-04
feat(validation-gen): Add declarative validation support for ResourceClaim/v1

yongruilin:
- Add tag to doc.go
- Generate code
- Add test scaffolding

This is generated by Gemini-CLI custom command


aaron-prindle:
LGTM

Currently this PR only adds this for the root "/" resource, the Gemini command might want to be able to support adding this for subresource though.  Just wanted to suggest that as currently that might be a gap for the Gemini command

yongruilin:
> LGTM
> 
> Currently this PR only adds this for the root "/" resource (in line with what we planned for v1.35 as ResourceClaimStatus is a separate task). Just wanted to note the Gemini command might want to be able to support wiring up subresources for other cases though. Just wanted to suggest that as currently that might be a gap for the Gemini command

The custom command actually has supported subresource, just I didn't request it in the prompt.

aaron-prindle:
When registering a type, we might also want the command to add the metrics label for the updated version of the metrics we are planning:

https://github.com/kubernetes/kubernetes/pull/133946/files

PR #182, opened 2025-09-09
refactor: ifEnabled/Disabled groups chained validations

yongruilin:
Key changes include:
   - Grouping multiple +k8s:ifEnabled and +k8s:ifDisabled rules for the same feature into a single conditional block.
   - Updating wrapper functions to accept multiple function generators.
   - Extending k8s:ifEnabled/k8s:ifDisabled to support all nilable types.

### DIFF
--- a/staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/tags/options/grouping/zz_generated.validations.go
+++ b/staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/tags/options/grouping/zz_generated.validations.go
@@ -63,6 +63,7 @@ func Validate_Struct(ctx context.Context, op operation.Operation, fldPath *field
 			}
 			// call field-attached validations
 			errs = append(errs, validate.IfOption(ctx, op, fldPath, obj, oldObj, "FeatureX", true, func(ctx context.Context, op operation.Operation, fldPath *field.Path, obj, oldObj []Item) field.ErrorList {
+				// lists with map semantics require unique keys
 				return validate.Unique(ctx, op, fldPath, obj, oldObj, func(a Item, b Item) bool { return a.Key == b.Key })
 			})...)
 			return
@@ -101,9 +102,11 @@ func Validate_Struct(ctx context.Context, op operation.Operation, fldPath *field
 			}
 			// call field-attached validations
 			errs = append(errs, validate.IfOption(ctx, op, fldPath, obj, oldObj, "FeatureX", false, func(ctx context.Context, op operation.Operation, fldPath *field.Path, obj, oldObj []Item) field.ErrorList {
+				// lists with map semantics require unique keys
 				return validate.Unique(ctx, op, fldPath, obj, oldObj, func(a Item, b Item) bool { return a.Key == b.Key })
 			})...)
 			errs = append(errs, validate.IfOption(ctx, op, fldPath, obj, oldObj, "FeatureX", true, func(ctx context.Context, op operation.Operation, fldPath *field.Path, obj, oldObj []Item) field.ErrorList {
+				// lists with map semantics require unique keys
 				return validate.RelaxUnique(ctx, op, fldPath, obj, oldObj, func(a Item, b Item) bool { return a.Key == b.Key })
 			})...)
 			return
--- a/staging/src/k8s.io/code-generator/cmd/validation-gen/validation.go
+++ b/staging/src/k8s.io/code-generator/cmd/validation-gen/validation.go
@@ -1490,6 +1490,9 @@ func toGolangSourceDataLiteral(sw *generator.SnippetWriter, c *generator.Context
 				sw.Do("    fldPath *$.field.Path|raw$, ", targs)
 				sw.Do("    obj, oldObj $.objTypePfx$$.objType|raw$ ", targs)
 				sw.Do(") $.field.ErrorList|raw$ {\n", targs)
+				for _, comment := range v.Functions[0].Comments {
+					sw.Do("// $.$\n", comment)
+				}
 				sw.Do("return ", nil)
 				emitFunctionCall(sw, c, v.Functions[0], "ctx", "op", "fldPath", "obj", "oldObj")
 				sw.Do("\n}", targs)


yongruilin:
@thockin This is a prerequisite for the possible solution to enable the listmap uniqueness for CSR. 
This PR's implementation need the tags to be grouped together. So the annotation for condition would looks like this
```go
	// +k8s:optional
	// +k8s:ifEnabled(CSRSemiStrictConditionDups)=+k8s:listType=map
	// +k8s:ifEnabled(CSRSemiStrictConditionDups)=+k8s:listMapKey=type
	// +k8s:ifEnabled(CSRSemiStrictConditionDups)=+k8s:item(type: "Approved")=+k8s:zeroOrOneOfMember
	// +k8s:ifEnabled(CSRSemiStrictConditionDups)=+k8s:item(type: "Denied")=+k8s:zeroOrOneOfMember
	// +k8s:ifDisabled(CSRSemiStrictConditionDups)=+k8s:listType=map
	// +k8s:ifDisabled(CSRSemiStrictConditionDups)=+k8s:listMapKey=type
	// +k8s:ifDisabled(CSRSemiStrictConditionDups)=+k8s:item(type: "Approved")=+k8s:zeroOrOneOfMember
	// +k8s:ifDisabled(CSRSemiStrictConditionDups)=+k8s:item(type: "Denied")=+k8s:zeroOrOneOfMember
        // +k8s:ifDisabled(CSRSemiStrictConditionDups)=+k8s:relaxUniquenessOnUpdateCondidtions
       Conditions []CertificateSigningRequestCondition
```
`CSRSemiStrictConditionDups` is the featuregate that the handwritten would use the more strict ratcheting logic that align with our expected pattern.

thockin:
Capturing some notes from chat:

This is (ab)using the field path to make 2 different "fields" and `ExtractValidations()` does something unexpected (to me) which is to call `FieldValidator`s for that non-field field.  My instinct is that this is a bug, it just happens to work in your favor.   Why a bug?  Well, we will end up calling all `FieldValidator`s on this virtual field, possibly BEFORE all the non-conditional `TagValidators` are run.  That makes any sort of collaboration between non-conditional tags and conditional tags impossible, and breaks our promises about ordering of validators.

IOW, something like:

```
+k8s:listType=map
+k8s:listMapKey=field1
+k8s:ifEnabled(BiggerKeys)=+k8s:listMapKey=field1
```

...could not work because you could get:

```
+k8s:listMapKey=field1
+k8s:ifEnabled(BiggerKeys)=+k8s:listMapKey=field2
+k8s:listType=map
```

...unless we assert some ordering between related tags which means users have to know which tags are related.  Blech

Ideally conditions are orthogonal to tags, orthogonal, but practically they are not.  We know things like `optional` do not compose with conditions. We know `listMapKey` does not either.

Joe changed it recently, maybe it was better to handle options as "special" ?

What would "special" look like for

```
+k8s:listType=map
+k8s:listMapKey=field1
+k8s:ifEnabled(BiggerKeys)=+k8s:relaxUniqueness
```

`listMetadata` would have to hold something like:

```
    uniquenessOpts []func(opts *uniquenessOpt)
```

and we would emit code to pass those opts to `validate.Unique()` like:

```
return validate.Unique(ctx, op, fldPath, obj, oldObj, func(a Item, b Item) bool { return a.Key1 == b.Key1 }, validate.Foo, validate.Bar, validate.Qux)
```

This solves for this ONE specific case, but basically any tag that isn't just a tag-chain will have to handle options explicitly.  Maybe not SO bad?

OR.....

1) we have a tag which disables uniqueness checking entirely
2) write manual validation code which handles this 
3) revisit this if we need it more than a few times

```
+k8s:listType=map
+k8s:listMapKey=field1
+k8s:unique=override // see manual validation code
```

It doesn't satisfy in the same way as making all tags truly conditional but that's a lot of work for relatively few cases?

I will keep chewing on alternative ideas or ways to make this PR work more suitably.

thockin:
Another idea: We could add a whole concept of "vfield" and "VFieldValidator".    `if` tag would change the context to ScopeVField when calling `listMapKey` or `relaxUniqueness`.  listMetadata would key on the VField path which is the real path + something extra.  When it came time to emit for list, we would have emit both the field and vfield functions, and the latter would have to carry information about the conditonal.

I can kind of see how it would go but it feels pretty complicated..

yongruilin:
I created https://github.com/jpbetz/validation-gen/pull/188 with unique=override

There are some other naming options (essentially it is disabling uniqueness in generated code):
- `k8s:customUnique` or `k8s:customUniqueness : I think it sounds like a plan.
- `k8s:omitDefaultUniqueness` : A bit too long?
- `k8s:handwrittenUnique`
- `k8s:NoUnique` : i don't like this one, as it is not accurate as handwritten may have uniqueness check

pkg/apis/apiserverinternal/validation/validation_test.go
```diff
@@ -430,13 +430,13 @@ func TestValidateStorageVersionName(t *testing.T) {
 		expectedErr: "",
 	}, {
 		name:        strings.Repeat("x", 254) + ".tokenreviews",
-		expectedErr: `the group segment must be no more than 253 characters`,
+		expectedErr: `the group segment must be no more than 253 bytes`,
```
aaron-prindle: Changing "characters" to "bytes" is technically correct for Go strings. Just a note to ensure we are comfortable that no downstream clients (e.g., UI or CLI parsers) strictly rely on the word "characters" in these error messages.

staging/src/k8s.io/apimachinery/pkg/api/validate/strfmt.go
```diff
@@ -152,3 +154,26 @@ func UUID[T ~string](_ context.Context, op operation.Operation, fldPath *field.P
...
+	if len(val) > 253 {
+		allErrs = append(allErrs, field.TooLong(fldPath, val, 253))
+	}
+	parts := strings.Split(val, "/")
+	for i, part := range parts {
+		if len(part) == 0 {
+			allErrs = append(allErrs, field.Invalid(fldPath, val, fmt.Sprintf("segment %d: must not be empty", i)))
+			continue
+		}
+		// Note that we are overwriting the origin from the underlying LongName validation.
```
aaron-prindle: Using `PrefixDetail` here is a nice touch. It gives necessary context (segment index) to the generic `LongName` error message without reimplementing the logic.

pkg/apis/batch/validation/validation.go
```diff
@@ -150,7 +152,7 @@ func validateGeneratedSelector(obj *batch.Job, validateBatchLabels bool) field.E
 // ValidateJob validates a Job and returns an ErrorList with any errors.
 func ValidateJob(job *batch.Job, opts JobValidationOptions) field.ErrorList {
 	// Jobs and rcs have the same name validation
-	allErrs := apivalidation.ValidateObjectMeta(&job.ObjectMeta, true, apivalidation.ValidateReplicationControllerName, field.NewPath("metadata"))
+	allErrs := apivalidation.ValidateObjectMeta(&job.ObjectMeta, true, apimachineryapivalidation.NameIsDNSSubdomain, field.NewPath("metadata"))
```
aaron-prindle: Good cleanup here replacing the resource-specific name validation with the generic `apimachineryapivalidation.NameIsDNSSubdomain`.

pkg/registry/core/replicationcontroller/strategy.go
```diff
@@ -177,10 +177,7 @@ func (rcStrategy) ValidateUpdate(ctx context.Context, obj, old runtime.Object) f
 	newRc := obj.(*api.ReplicationController)
 
 	opts := pod.GetValidationOptionsFromPodTemplate(newRc.Spec.Template, oldRc.Spec.Template)
-	// This should be fixed to avoid the redundant calls, but carefully.
-	validationErrorList := corevalidation.ValidateReplicationController(newRc, opts)
-	updateErrorList := corevalidation.ValidateReplicationControllerUpdate(newRc, oldRc, opts)
-	errs := append(validationErrorList, updateErrorList...)
+	errs := corevalidation.ValidateReplicationControllerUpdate(newRc, oldRc, opts)
```
aaron-prindle: Removing the redundant `ValidateReplicationController` call is a good optimization, assuming `ValidateReplicationControllerUpdate` covers all necessary static validation checks or that they are immutable.

staging/src/k8s.io/apimachinery/pkg/api/validate/options.go
```diff
@@ -25,8 +25,8 @@ import (
 
 // IfOption conditionally evaluates a validation function. If the option and enabled are both true the validator
 // is called. If the option and enabled are both false the validator is called. Otherwise, the validator is not called.
-func IfOption[T any](ctx context.Context, op operation.Operation, fldPath *field.Path, value, oldValue *T,
-	optionName string, enabled bool, validator func(context.Context, operation.Operation, *field.Path, *T, *T) field.ErrorList,
+func IfOption[T any](ctx context.Context, op operation.Operation, fldPath *field.Path, value, oldValue T,
```
aaron-prindle: `IfOption` seems like a key primitive for supporting feature-gated validations.

staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/cohorts/doc.go
```diff
@@ -0,0 +1,55 @@
...
+
+var localSchemeBuilder = testscheme.New()
+
+// Note: the following are out of order on purpose.
+// They should be emitted as:
+//   - T {ShortCircuit,Regular}
+//   - T c2 {ShortCircuit,Regular}
+//   - T c1 {ShortCircuit,Regular}
+//
+// +k8s:validateFalse(cohort: "c2")="type T c2 Regular"
```
aaron-prindle: The test case here clearly demonstrates the "cohort" functionality for grouping validations (`c1`, `c2`) and flags (`ShortCircuit`). This is helpful for understanding the feature.

staging/src/k8s.io/apimachinery/pkg/util/validation/field/error_matcher.go
```diff
@@ -138,6 +138,13 @@ func (m ErrorMatcher) ByValue() ErrorMatcher {
 }
 
 // ByOrigin returns a derived ErrorMatcher which also matches by the origin.
+// When this is used and an origin is set in the error, the matcher will
```
aaron-prindle: The `ByOrigin()` matcher is a nice enhancement for testing, allowing us to verify that an error came from a specific validation layer (e.g. "format=k8s-ip").

pkg/registry/resource/resourceclaim/strategy.go
```diff
@@ -184,8 +206,25 @@ func (r *resourceclaimStatusStrategy) ValidateUpdate(ctx context.Context, obj, o
...
-	allErrs := resourceutils.AuthorizedForAdminStatus(ctx, newAllocationResult, oldAllocationResult, newClaim.Namespace, r.nsClient)
-	return append(allErrs, validation.ValidateResourceClaimStatusUpdate(newClaim, oldClaim)...)
+	errs := resourceutils.AuthorizedForAdminStatus(ctx, newAllocationResult, oldAllocationResult, newClaim.Namespace, r.nsClient)
+	errs = append(errs, validation.ValidateResourceClaimStatusUpdate(newClaim, oldClaim)...)
+
+	// If DeclarativeValidation feature gate is enabled, also run declarative validation
+	if utilfeature.DefaultFeatureGate.Enabled(features.DeclarativeValidation) {
+		// Determine if takeover is enabled
+		takeover := utilfeature.DefaultFeatureGate.Enabled(features.DeclarativeValidationTakeover)
+
```
aaron-prindle: The use of `CompareDeclarativeErrorsAndEmitMismatches` is excellent for observing the rollout of the new validation logic before it becomes authoritative (takeover).

Issue #63, opened 2025-05-18
Discussion: How should validation-gen handle two "flavors" of immutable


aaron-prindle:
This issue is to discuss and build consensus on how to support the two types of common immutability patterns in k/k. The goal is to provide dedicated, simple-to-use validation tags (e.g., `+k8s:immutable`) that can handle these patterns without requiring users to write complex conditional CEL expressions or Go code involving `oldSelf`.

1.  **Create-Time Immutability ("Cannot Change After Create"):**

    -   **Definition:** A field's value can be set (or must be set) when an object is first created, but it cannot be changed in any subsequent update operation. If the field is optional and not set at creation, it often cannot be set later either (remaining at its zero/default value).
    -   **Current Practice:** This is often implemented in API validation functions (e.g., `ValidateFooUpdate` checking `oldObj.Spec.Field != newObj.Spec.Field`).
    -   **Kubernetes Examples:**
        -   `metadata.name`: The name of an object is set at creation and cannot be changed.
        -   `metadata.namespace`: The namespace of an object is set at creation and cannot be changed.
        -   `PersistentVolume.spec.capacity`: The storage capacity of a PV is defined at provisioning and is immutable.
        -   `StorageClass.provisioner`: The provisioner for a StorageClass is set at creation and cannot be changed as it defines the fundamental behavior of the StorageClass.
        -   `EndpointSlice.addressType`: This must be specified when an EndpointSlice is created and cannot be changed afterward.

undefined

1.  **Set-Once Immutability ("Cannot Change Once Set"):**

    -   **Definition:** A field can be left unset (or at its zero/default value) during object creation. It can be set to a non-default value in a subsequent update. However, once it has been set to a non-default value, it cannot be changed to a *different* non-default value, nor can it (typically) be unset back to its zero/default value.
    -   **Current Practice:** This requires more complex logic, often checking `oldObj.Spec.Field` being empty/nil and `newObj.Spec.Field` being non-empty, and then ensuring `oldObj.Spec.Field == newObj.Spec.Field` if `oldObj.Spec.Field` was already set.
    -   **Kubernetes Examples:**
        -   `Pod.spec.nodeName`: A Pod is created without a `nodeName`. The scheduler then assigns a node to it by setting this field. Once set, the Pod is bound to that node and `nodeName` cannot be changed.
[...]

PR #186, opened 2025-09-15
feat(validation-gen): add +k8s:update tag and associated tests

aaron-prindle:
This PR adds a new +k8s:update tag with support for non-lists/non-maps via with the allowable update constraints - NoSet, NoModify, NoUnset.  

```go
// +k8s:update=NoSet
// +k8s:update=NoModify
// +k8s:update=NoSet
```
List/map support will be done in a followup PR that adds NoAddItem and NoRemoveItem constraints

As a requirement for +k8s:immutable and +k8s:update to properly stop a fields from the transition set -> unset, ShortCircuiting needs to be updated so that all short-circuit tags are run and then non-short-circuit tags are short circuiting if any short circuit tag errors.  Without this, the +k8s:optional check would stop the +k8s:update=NoUnset and +k8s:immutable tags from running which is not correct

+k8s:immutable is minorly changed in this PR adding the ShortCircuit tag.  In the future, +k8s:immutable will be modified to have the correct semantics based on the field it is added to (non-lists/maps, lists/maps, and list/map items)

Related issue: https://github.com/jpbetz/validation-gen/issues/63


staging/src/k8s.io/code-generator/cmd/validation-gen/validators/update.go
```diff
@@ -0,0 +1,228 @@
...
+	return nil
+}
+
+func (utc updateTagCollector) Docs() TagDoc {
+	return TagDoc{
+		Tag:          utc.TagName(),
+		Scopes:       utc.ValidScopes().UnsortedList(),
+		PayloadsType: codetags.ValueTypeString,
+		Description: "Provides fine-grained control over field update operations. " +
+			"Only for non-list/map fields: NoSet (prevents unset->set), NoUnset (prevents set->unset), " +
```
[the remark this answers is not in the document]
aaron-prindle: This no longer supports a comma-separated list of constraints so removed that part.  I have updated the description to have the full context for types now and explain the future work:

		Description: "Provides constraints on the allowed update operations of a field. " +
			"Currently supports non-list and non-map fields only. " +
			"Constraints: NoSet (prevents unset->set transitions), NoUnset (prevents set->unset transitions), " +
			"NoModify (prevents value changes but allows set/unset transitions). " +
			"Multiple constraints can be specified using multiple tags. " +
			"For non-pointer structs, NoSet and NoUnset have no effect as these fields cannot be unset. " +
			"Future support planned for lists/maps with NoAddItem and NoRemoveItem constraints. " +
			"Examples: +k8s:update=NoModify +k8s:update=NoUnset for set-once fields; " +
			"+k8s:update=NoSet for fields that must be set at creation or never.",

staging/src/k8s.io/apimachinery/pkg/api/validate/update.go
```diff
@@ -0,0 +1,175 @@
...
+		return field.ErrorList{
+			field.Forbidden(fldPath, "field cannot be set once created"),
+		}
+	}
+
+	return nil
+}
+
+// NoSetPointer verifies that a pointer field cannot be set (transition from nil to non-nil).
+func NoSetPointer[T any](_ context.Context, op operation.Operation, fldPath *field.Path, value, oldValue *T) field.ErrorList {
```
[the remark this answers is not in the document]
aaron-prindle: I have modified validate/update.go from the comment here: https://github.com/jpbetz/validation-gen/pull/186#discussion_r2352984896 and now the validate/*.go methods are more aligned, see below.  Happy to change the naming/format of methods there to make them align better:

validate/update.go
- UpdateValueByCompare
- UpdatePointer
- UpdateValueByReflect
- UpdateStruct

validate/immutable.go
- ImmutableByCompare
- ImmutableByReflect

aaron-prindle:
@jpbetz  - PR is rebased now, CI is green, and all initial review comments are addressed. PTAL when you get a chance, thanks.

staging/src/k8s.io/apimachinery/pkg/api/validate/update.go
```diff
@@ -0,0 +1,153 @@
...
+			if oldValue == nil && value != nil {
+				errs = append(errs, field.Forbidden(fldPath, "field cannot be set once created"))
+			}
+		case NoUnset:
+			if oldValue != nil && value == nil {
+				errs = append(errs, field.Forbidden(fldPath, "field cannot be cleared once set"))
+			}
+		case NoModify:
+			// Allow transitions between set/unset
+			if oldValue != nil && value != nil && !equality.Semantic.DeepEqual(value, oldValue) {
```
thockin, line 89: Why no UpdatePointerByCompare option?

staging/src/k8s.io/apimachinery/pkg/api/validate/update.go
```diff
@@ -0,0 +1,153 @@
...
+				errs = append(errs, field.Forbidden(fldPath, "field cannot be modified once set"))
+			}
+		}
+	}
+
+	return errs
+}
+
+// UpdateValueByReflect verifies update constraints for non-comparable value types using reflection.
+func UpdateValueByReflect[T any](_ context.Context, op operation.Operation, fldPath *field.Path, value, oldValue *T, constraints ...UpdateConstraint) field.ErrorList {
```
thockin, line 99: Pretty minor, but I see you followed the Immutable pattern of having 2 function entry-points rather than the MatchFunc pattern used in other places.  I'm afraid we will end up needing all of `Update{Value,Pointer,Struct,List,Map}By{Compare,Reflect}()`.  We should probably pick one pattern and stick with it?

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/immutable.go
```diff
@@ -59,9 +59,12 @@ func (immutableTagValidator) GetValidations(context Context, _ codetags.Tag) (Va
 		// pointer fields, which are directly comparable but not what we need.
 		//
 		// Note: This compares the pointee, not the pointer itself.
-		result.AddFunction(Function(immutableTagName, DefaultFlags, immutableCompareValidator))
+
+		// Use ShortCircuit flag so immutable runs in the same group as +k8s:optional
+		result.AddFunction(Function(immutableTagName, ShortCircuit, immutableCompareValidator))
```
thockin, line 64: Should we consider gutting the `immutable` implementation and make that tag just set all the `No*` update constraints?  Or is it worthwhile to have a different error?

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/update.go
```diff
@@ -0,0 +1,225 @@
...
+	t := util.NonPointer(util.NativeType(context.Type))
+	if t.Kind == types.Slice || t.Kind == types.Map {
+		// TODO: add support for list and map fields
+		return Validations{}, fmt.Errorf("update constraints are currently not supported on list or map fields")
+	}
+
+	return ufv.generateValidation(context, constraints)
+}
+
+func (ufv updateFieldValidator) generateValidation(context Context, constraints []validate.UpdateConstraint) (Validations, error) {
```
thockin, line 182: if constraints is empty, should we just return and not emit anything at all?

thockin (commented):
Looks nice !  Just some smaller questions, sorry I am late.

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/immutable.go
aaron-prindle: Addressed this in:
https://github.com/jpbetz/validation-gen/pull/203

With the change to make validate/immutable.go essentially always error such logic does not seem necessary.

staging/src/k8s.io/apimachinery/pkg/api/validate/update.go
aaron-prindle: Added in new PR here: https://github.com/jpbetz/validation-gen/pull/203

PR #188, opened 2025-09-16
[POC] Add unique=override to enable listmap uniqueness

yongruilin:


staging/src/k8s.io/code-generator/cmd/validation-gen/validators/list.go
```diff
@@ -363,6 +373,12 @@ func (lv listValidator) GetValidations(context Context) (Validations, error) {
 		return Validations{}, err
 	}
 
+	if lm.uniqueOverride {
+		// Uniqueness validation is disabled in generated validation for this list.
+		// It would defer to handwritten validation to check the uniqueness.
+		return Validations{}, nil
```
thockin, line 379: Could return `result.AddComment("uniquess checking is overridden")` or something.

thockin (commented):
As a POC this looks really good.  I didn't think I would like it using +unique and I was right :)

Ideas for a better name, or good arguments why it SHOULD use +unique?

staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/tags/optional/zero_defaults/zz_generated.validations.go
```diff
@@ -57,11 +57,6 @@ func Validate_Struct(ctx context.Context, op operation.Operation, fldPath *field
 	errs = append(errs,
 		func(fldPath *field.Path, obj, oldObj *string) (errs field.ErrorList) {
 			// optional value-type fields with zero-value defaults are purely documentation
-			// don't revalidate unchanged data
```
thockin, line 60: What made this change?

thockin: NM, I get it.

thockin (commented):
I am LGTM on this - let's give time for Joe or Aaron to look at it, but if no comments , you can self-merge :)

staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/tags/optional/zero_defaults/zz_generated.validations.go
aaron-prindle: To confirm my understanding, this change is caused because of the updated:

```go
				if len(validations.Functions) > 0 {
					emitRatchetingCheck(c, fld.childType, bufsw)
					fldRatchetingChecked = true
					bufsw.Do("// call field-attached validations\n", nil)
					emitCallsToValidators(c, validations.Functions, bufsw)
				}
```

+k8s:optional is not part of validations.Functions but if there was a validation on this field then the ratcheting check would be generated

yongruilin: Yes, basically this change would avoid useless ratcheting check.

PR #195, opened 2025-09-18
feat(validation-gen): migrate ResourceClaimStatus.Allocation NoModify validation to declarative validation

aaron-prindle:
Migrate +k8s:update=NoModify on ResourceClaimStatus.Allocation

The hand-written validation logic for ResourceClaimStatus.Allocation allows for setting and unsetting but not modifying:
```go
	// Updates to a populated status.Allocation are not allowed.
	// Unmodified fields don't need to be validated again and,
	// in this particular case, must not be validated again because
	// validation for new results is tighter than it was before.
	if oldStatus.Allocation != nil && status.Allocation != nil {
		allErrs = append(allErrs, apimachineryvalidation.ValidateImmutableField(status.Allocation, oldStatus.Allocation, fldPath.Child("allocation"))...)
	} else if status.Allocation != nil {
		allErrs = append(allErrs, validateAllocationResult(status.Allocation, fldPath.Child("allocation"), requestNames, false)...)
	}
```

This PR also updates the +k8s:update tag to:
- use Invalid errors (vs Forbidden errors) to match with the hand-written immutable validation methos
- adds Origin information to +k8s:update that was omitted initially  

This PR also updates our Github CI workflow YAML to run tests in the:
- resourceclaim
- deviceclass
dirs to run the declarative_validation_test.go files for those


pkg/registry/resource/resourceclaim/declarative_validation_test.go
```diff
@@ -201,6 +201,49 @@ func TestValidateStatusUpdateForDeclarative(t *testing.T) {
...
 			},
 		},
+		"valid status.allocation unchanged": {
+			old:    mkResourceClaimWithStatus(),
+			update: mkResourceClaimWithStatus(),
+		},
+		"valid status.allocation set from nil": {
+			old:    mkValidResourceClaim(),
+			update: mkResourceClaimWithStatus(),
+		},
```
jpbetz, line 211: unset is also allowed, right? We should test that.  Seeing handwritten and DV agree on that feels essential here.

aaron-prindle: Ah good catch, I initially had this has NoUnset and NoModify and in changing to NoModify I dropped this test but as mentioned it should still be there to ensure Unset is allowed for both.  I've added this test case now:

```go
		"valid status.allocation cleared (Unset is allowed)": {
			old:    mkResourceClaimWithStatus(),
			update: mkValidResourceClaim(),
		},
```

jpbetz:
LGTM

PR #196, opened 2025-09-18
Simplify testing of Scale API

### LINKED ISSUE #192: declarative_validation_test for Scale doesn't compare between handwritten and generated
https://github.com/jpbetz/validation-gen/blob/06bc106b3bc4b3a1d96d41924bb3adc40580b10c/pkg/apis/autoscaling/validation/declarative_validation_test.go#L58-L63

It only always run `rest.ValidateDeclaratively` either the featuregates are enabled or disabled. So it actually doesn't compare the errors between handwritten and generated code.

jpbetz:
Fixes #192

### DIFF
--- a/pkg/apis/autoscaling/validation/declarative_validation_test.go
+++ b/pkg/apis/autoscaling/validation/declarative_validation_test.go
@@ -17,11 +17,9 @@ limitations under the License.
 package validation
 
 import (
-	"context"
 	"testing"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	genericapirequest "k8s.io/apiserver/pkg/endpoints/request"
 	"k8s.io/apiserver/pkg/registry/rest"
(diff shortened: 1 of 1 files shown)


pkg/apis/autoscaling/validation/declarative_validation_test.go
```diff
@@ -56,9 +47,11 @@ func TestScaleDeclarativeValidation(t *testing.T) {
...
 	for k, tc := range testCases {
 		t.Run(k, func(t *testing.T) {
-			apitesting.VerifyValidationEquivalence(t, ctx, &tc.input, func(ctx context.Context, obj runtime.Object) field.ErrorList {
-				return rest.ValidateDeclaratively(ctx, legacyscheme.Scheme, obj)
-			}, tc.expectedErrs, "scale")
+			// The Scale API has no handwritten validation to compare against. So we simply cross-test all the versions
+			// of the Scale subresource.
+			// All resources that have a scale subresource are expected to test Scale validation against any handwritten
+			// validation code defined on that resource.
+			apitesting.VerifyVersionedValidationEquivalence(t, &tc.input, nil, "scale")
```
yongruilin, line 54: In this case, the `tc.expectedErrs` is no longer used. `VerifyVersionedValidationEquivalence` only checks equivalence without comparing it with expected errors. We can drop the expectedErrs in this test? Maybe even rename the func to `TestVersionedScaleDeclarativeValidationEquivalence`

yongruilin: LGTM, I will merge it after CI passed

PR #204, opened 2025-09-30
fix: make +k8s:immutable tag have Invalid error (vs Forbidden) to better match hand-written usage

aaron-prindle:
This PR changes +k8s:immutable to output an Invalid error (vs Forbidden) .  This is to align with the kubernetes `ValidateImmutableField`:

https://github.com/kubernetes/kubernetes/blob/8f372d2ac06c78707a92c1e88e5e0bb1fc5b59aa/staging/src/k8s.io/apimachinery/pkg/api/validation/objectmeta.go#L130-L137

```go
// ValidateImmutableField validates the new value and the old value are deeply equal.
func ValidateImmutableField(newVal, oldVal interface{}, fldPath *field.Path) field.ErrorList {
	allErrs := field.ErrorList{}
	if !apiequality.Semantic.DeepEqual(oldVal, newVal) {
		allErrs = append(allErrs, field.Invalid(fldPath, newVal, FieldImmutableErrorMsg))
	}
	return allErrs
}
```

### DIFF
--- a/staging/src/k8s.io/apimachinery/pkg/api/validate/immutable.go
+++ b/staging/src/k8s.io/apimachinery/pkg/api/validate/immutable.go
@@ -35,6 +35,6 @@ func Immutable[T any](_ context.Context, op operation.Operation, fldPath *field.
 		return nil
 	}
 	return field.ErrorList{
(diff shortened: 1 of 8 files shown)


yongruilin (approved):
LGTM.
It is also consistent with what we have in `validate/update.go`

PR #207, opened 2025-10-06
Allow pointers for list keys

lalitc375:
This PR allows using pointers as keys in `listType=map` and `listType=set` validations.

#### What type of PR is this?

/kind feature

This change extends the `validation-gen` tool to support pointer types for `+k8s:listMapKey`. Previously, only non-pointer primitive types were allowed as list map keys.

This enables validation for list types where the map keys are pointers. The generated validation code now correctly handles `nil` checks and dereferences pointers for key comparisons in uniqueness checks and for `+k8s:item` rules.

The changes include:
-   Updates to the `list` and `item` validators to handle pointer types.
-   The `makeListMapMatchFunc` now generates code that correctly compares pointer keys, accounting for `nil` values.
-   The `item` validator is updated to correctly build match conditions for pointer types in `+k8s:item` rules.
-   Added comprehensive test cases for single and multiple pointer keys, and mixed pointer/non-pointer keys for both `item` and `listmap` validation.

N/A

#### Special notes for your reviewer:

NONE

#### Does this PR introduce a user-facing change?
```release-note
NONE
```

### DIFF
--- a/pkg/apis/batch/validation/validation.go
+++ b/pkg/apis/batch/validation/validation.go
@@ -26,6 +26,7 @@ import (
 	"github.com/robfig/cron/v3"
 
 	apiequality "k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/api/validate/content"
 	apimachineryapivalidation "k8s.io/apimachinery/pkg/api/validation"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	unversionedvalidation "k8s.io/apimachinery/pkg/apis/meta/v1/validation"
--- a/staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/tags/item/multiple_keys/doc.go
+++ b/staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/tags/item/multiple_keys/doc.go
@@ -41,6 +41,19 @@ type Struct struct {
 	// +k8s:listMapKey=boolKey
 	// +k8s:item(boolKey: true, stringKey: "target", intKey: 42)=+k8s:validateFalse="item OutOfOrder[boolKey=42,stringKey=target,intKey=42]"
 	OutOfOrder []Item `json:"outOfOrder"`
+
+	// +k8s:listType=map
+	// +k8s:listMapKey=stringKey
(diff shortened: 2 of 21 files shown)


staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/tags/item/multiple_keys/doc.go
```diff
@@ -41,6 +41,19 @@ type Struct struct {
 	// +k8s:listMapKey=boolKey
 	// +k8s:item(boolKey: true, stringKey: "target", intKey: 42)=+k8s:validateFalse="item OutOfOrder[boolKey=42,stringKey=target,intKey=42]"
 	OutOfOrder []Item `json:"outOfOrder"`
+
```
jpbetz, line 44: Can we also add output tests for `unique`, `eachKey` and `eachVal`?

pkg/apis/batch/validation/validation.go
```diff
@@ -26,6 +26,7 @@ import (
 	"github.com/robfig/cron/v3"
 
 	apiequality "k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/api/validate/content"
```
jpbetz, line 29: I don't see where this added import gets used. Is this correct?

jpbetz (commented):
Implementation looks correct. Just one request for expanded test coverage then LGTM.

staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/tags/item/multiple_keys/doc.go
lalitc375: sure.

yongruilin: `makeListMapMatchFunc` is used in validators/each.go to construct equiv func for EachSliceVal's ratcheting check. Do we need to have test cases for that?

lalitc375: I have added ratcheting test for this combination.

PR #211, opened 2025-10-31
(feat): add maxProperties declarative validation tag

### LINKED ISSUE #208: Add +k8s:maxProperties tag
Issue tracks adding a +k8s:maxProperties tag. This tag is used to limit the number of properties (keys) in a
  map. It should function similarly to other limit tags like +k8s:maxItems or +k8s:maxLength.

  Expected implementation locations:

  validation method logic:
   * staging/src/k8s.io/apimachinery/pkg/api/validate/limits.go (Add MaxProperties function)
   * staging/src/k8s.io/apimachinery/pkg/api/validate/limits_test.go

  validators logic:
   * staging/src/k8s.io/code-generator/cmd/validation-gen/validators/limits.go (Update to parse the new tag)

  validation tag test:
   * staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/tags/maxproperties (New directory for
     generation tests)

  maxProperties is semantically identical to the maxProperties Open API validation.

  Why is this needed?

  This tag is needed as maxProperties is a part of the Open API validation that Declarative Validation needs to
  support, and k/k types use this validation to bound the size of maps.

everettraven:
#### What type of PR is this?

/kind feature

Adds a new declarative validation tag, `k8s:maxProperties`, that ensures that maps do not contain more than X properties.

Fixes #208


staging/src/k8s.io/apimachinery/pkg/api/validate/limits.go
```diff
@@ -45,6 +45,18 @@ func MaxItems[T any](_ context.Context, _ operation.Operation, fldPath *field.Pa
 	return nil
 }
 
+// MaxProperties verifies that the specified map has no more than max keys.
+func MaxProperties[K comparable, V any](_ context.Context, _ operation.Operation, fldPath *field.Path, value, _ map[K]V, max int) field.ErrorList {
```
everettraven: Open question for reviewers - should this be made ratcheting aware?

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/limits.go
```diff
@@ -52,9 +54,7 @@ func (maxLengthTagValidator) ValidScopes() sets.Set[Scope] {
 	return maxLengthTagValidScopes
 }
 
-var (
-	maxLengthValidator = types.Name{Package: libValidationPkg, Name: "MaxLength"}
-)
+var maxLengthValidator = types.Name{Package: libValidationPkg, Name: "MaxLength"}
```
everettraven: Automated format change that happened on my end. If reviewers would prefer, I can revert formatting changes like this.

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/limits.go
```diff
@@ -149,6 +147,60 @@ func (mitv maxItemsTagValidator) Docs() TagDoc {
...
+
+var maxPropertiesValidator = types.Name{Package: libValidationPkg, Name: "MaxProperties"}
+
+func (maxPropertiesTagValidator) GetValidations(context Context, tag codetags.Tag) (Validations, error) {
+	var result Validations
+
+	// NOTE: pointers to maps are not supported, so we should never see a pointer here.
+	if t := util.NativeType(context.Type); t.Kind != types.Map && t.Key.Kind != types.Builtin && t.Key.Name.Name != "string" {
+		return Validations{}, fmt.Errorf("can only be used on map types with string-based keys (%s)", rootTypeString(context.Type, t))
+	}
```
everettraven: My understanding is that for APIs that have a field with a map, the keys must be a string-based type (can be a type alias to a key).

If this isn't accurate, or is not the right place to enforce this, let me know and I can do a simple map type enforcement.

staging/src/k8s.io/apimachinery/pkg/api/validate/limits.go
aaron-prindle: > Open question for reviewers - does this need anything special to be made ratcheting aware?

For this tag, nothing additionally needs to be done for ratcheting.  This should be ratcheting aware using only the default ratcheting that is generated over fields in a type.  Here is an example of the default ratcheting which runs before validations:

https://github.com/kubernetes/kubernetes/blob/d2e1e3b6bc9fd5330bb9e2671c026634b18106bb/staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/tags/maxitems/slice_of_primitive/zz_generated.validations.go#L60-L63

If a tag is "cross-field" and the validation relies on information from multiple fields (+k8s:unionMember, etc) then the validate/*.go logic does need to account for ratcheting (not the case here)

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/limits.go
```diff
@@ -149,6 +147,60 @@ func (mitv maxItemsTagValidator) Docs() TagDoc {
...
+
+	intVal, err := strconv.Atoi(tag.Value)
+	if err != nil {
+		return result, fmt.Errorf("failed to parse tag payload as int: %w", err)
+	}
+	if intVal < 0 {
+		return result, fmt.Errorf("must be greater than or equal to zero")
+	}
+	// Note: maxProperties short-circuits other validations.
+	result.AddFunction(Function(maxPropertiesTagName, ShortCircuit, maxPropertiesValidator, intVal))
```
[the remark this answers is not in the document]
aaron-prindle: I was incorrect, we actually do want the short-circuit tag here.  No changes required, marking this thread resolved.

PR #219, opened 2025-11-20
Enforce +k8s:required or +k8s:optional for pointers with validations

lalitc375:
#### What type of PR is this?

/kind feature

This PR improves the validation logic in two ways:

1.  **Enforces explicit optionality for pointers:** It ensures that any pointer field with validation rules must be explicitly marked with either `+k8s:required` or `+k8s:optional`. This removes ambiguity and prevents validations from being silently ignored or causing un-intended failures. 

The changes in this PR consist of:

*   An update to `staging/src/k8s.io/code-generator/cmd/validation-gen/validation.go` to enforce the new rule for pointers.
*   Updates across various test files to add the `+k8s:optional` tag to pointer fields, aligning them with the new validation requirement.

N/A

### DIFF
--- a/staging/src/k8s.io/code-generator/cmd/validation-gen/validation.go
+++ b/staging/src/k8s.io/code-generator/cmd/validation-gen/validation.go
@@ -600,8 +600,6 @@ func (td *typeDiscoverer) discoverStruct(thisNode *typeNode, fldPath *field.Path
 
 	klog.V(5).InfoS("discoverStruct", "type", thisNode.valueType)
 
-	var requiredOrOptionalTags = []string{"k8s:required", "k8s:optional"}
-
 	// Discover into each field of this struct.
 	for _, memb := range thisNode.valueType.Members {
 		name := memb.Name
@@ -806,17 +804,8 @@ func (td *typeDiscoverer) discoverStruct(thisNode *typeNode, fldPath *field.Path
 		}
 
 		// Check for missing required/optional tags on pointers with validations.
-		if childType.Kind == types.Pointer {
-			hasTag := checkTags(memb.CommentLines, requiredOrOptionalTags)
-			if !hasTag {
-				hasVals := !child.fieldValidations.Empty()
-				if !hasVals && child.node != nil {
-					hasVals = nodeHasValidations(child.node)
-				}
-				if hasVals {
-					return fmt.Errorf("field %s is a pointer and has validations (or contains fields with validations) but is missing +k8s:required or +k8s:optional tag", childPath)
-				}
-			}
+		if err := td.checkPointerOptionality(child, memb.CommentLines, childPath); err != nil {
+			return err
 		}
 
 		fields = append(fields, child)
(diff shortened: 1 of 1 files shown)


staging/src/k8s.io/code-generator/cmd/validation-gen/validation.go
```diff
@@ -802,6 +805,20 @@ func (td *typeDiscoverer) discoverStruct(thisNode *typeNode, fldPath *field.Path
...
+				hasVals := !child.fieldValidations.Empty()
+				if !hasVals && child.node != nil {
+					hasVals = nodeHasValidations(child.node)
+				}
+				if hasVals {
+					return fmt.Errorf("field %s is a pointer and has validations (or contains fields with validations) but is missing +k8s:required or +k8s:optional tag", childPath)
+				}
+			}
+		}
+
```
jpbetz: The reason we're implementing this as a validation-gen error instead of a lint rule is because validation-gen's behavior is incorrect without the `+optional` tag?   Can we expand the comment to more clearly justify *why* this is a hard rule?

jpbetz:
cc @yongruilin
For linter vs. error in validation-gen topic

staging/src/k8s.io/code-generator/cmd/validation-gen/validation.go
```diff
@@ -875,16 +881,19 @@ func mkSymbolArgs(c *generator.Context, names []types.Name) generator.Args {
 	return args
 }
 
-// hasValidations checks whether the given typeNode has any
+// nodeHasValidations checks whether the given typeNode has any
 // validations, transitively.
-func (g *genValidations) hasValidations(n *typeNode) bool {
+func nodeHasValidations(n *typeNode) bool {
```
thockin: This can't work, I think.  See commit 641834d030ec51b351707f3d8058451e574f00f4

It used to be this way, and it turns out to be a problem for recursive types, which are not fully resolved  at the time this is called.  I specifically made it a method of `genValidation` to remove tyhe temptation to call it during discovery.

thockin (commented):
I don't think this belongs here and I don't know that it works anyway.

Why can't we extend the built-in lint?  What we call `lintComments` could probably become `lintField` and pass the field's `Type` to the rule.  That can look for pointers and look at the tags.

I also don't love that this "pollutes" all the tests with tags and generated code that they don't need, making it that much harder to find the signal in the noise.

staging/src/k8s.io/code-generator/cmd/validation-gen/validation.go
thockin: By all means look at it, but know that I spent a good amount of time trying to figure it out, and the only answer I found was "don't call hasValidations() in discovery".  

We have a lint pass, and it runs AFTER discovery, for a reason.

lalitc375:
> No, look in staging/src/k8s.io/code-generator/cmd/validation-gen/lint* -- that should be easy to extend.

If I am not making it up, there is [a plan](https://github.com/kubernetes-sigs/kube-api-linter/pulls?q=author%3Ayongruilin) to deprecate this linter and use KAL. It will be very hard in KAL.

PR #224, opened 2025-12-03
strategy.go : support declarative native validation

lalitc375:
This PR enables declarative-only validations to run even when the `DeclarativeValidation` feature gate is disabled. This is useful for cases where a validation is defined only declaratively and does not have an imperative counterpart.

#### What type of PR is this?

/kind feature

This change introduces a mechanism to support "declarative-only" validations. Previously, all declarative validations were skipped if the `DeclarativeValidation` feature gate was disabled. This PR modifies the validation logic to allow specific validations to be marked as "declarative-only," ensuring they are executed regardless of the feature gate's status.

This is achieved by:
- Adding a `WithContainsDeclarativelyOnly` option to mark a validation configuration as containing declarative-only rules.
- Updating `ValidateDeclarativelyWithMigrationChecks` to run if either the `DeclarativeValidation` feature gate is on, or if the configuration is marked as containing declarative-only rules.
- Separating declarative errors into mirrored and declarative-only errors, and ensuring the latter are always included in the final error list.

### Behavior Matrix

The following table summarizes the behavior of the validation logic based on the different combinations of the `DeclarativeValidation` and `DeclarativeValidationTakeover` feature gates and the `containsDeclarativeOnly` configuration option.

| `DeclarativeValidation` | `DeclarativeValidationTakeover` | `containsDeclarativeOnly` | Resulting Behavior |
| :--- | :--- | :--- | :--- |
| Disabled | (any) | `false` | Only imperative validation errors are returned. |
| Disabled | (any) | `true` | Imperative errors and "declarative-only" errors are returned. Mirrored declarative errors are ignored. |
| Enabled | Disabled | (any) | Imperative errors and "declarative-only" errors are returned. |
| Enabled | Enabled | `false` | Non-covered imperative errors and all declarative errors are returned. |
| Enabled | Enabled | `true` | Non-covered imperative errors, mirrored declarative errors, and "declarative-only" errors are returned. |

#### Special notes for your reviewer:

- The core logic change is in `ValidateDeclarativelyWithMigrationChecks` in `staging/src/k8s.io/apiserver/pkg/registry/rest/validate.go`.
- New tests have been added in `TestValidateDeclarativelyWithMigrationChecks` in `staging/src/k8s.io/apiserver/pkg/registry/rest/validate_test.go` to cover the new functionality.

#### Does this PR introduce a user-facing change?
```release-note
NONE
```

### DIFF
--- a/staging/src/k8s.io/apiserver/pkg/registry/rest/validate.go
+++ b/staging/src/k8s.io/apiserver/pkg/registry/rest/validate.go
@@ -75,24 +75,24 @@ func WithNormalizationRules(rules []field.NormalizationRule) ValidationConfig {
 	}
 }
 
-// WithContainsDeclarativelyOnly marks the validation configuration to indicate that it includes
+// WithDeclarativeNative marks the validation configuration to indicate that it includes
 // declarative validations that are defined *only* declaratively, lacking corresponding imperative validation.
 // When set, declarative validation is always executed regardless of feature gates. Errors marked as
-// declarative-only are separated from the full set and returned alongside imperative errors.
-func WithContainsDeclarativelyOnly() ValidationConfig {
+// declarative-native are separated from the full set and returned alongside imperative errors.
+func WithDeclarativeNative() ValidationConfig {
 	return func(config *validationConfigOption) {
-		config.containsDeclarativeOnly = true
+		config.containsDeclarativeNative = true
 	}
 }
 
 type validationConfigOption struct {
-	opType                  operation.Type
-	options                 []string
-	takeover                bool
-	subresourceGVKMapper    GroupVersionKindProvider
-	validationIdentifier    string
-	normalizationRules      []field.NormalizationRule
-	containsDeclarativeOnly bool
+	opType                    operation.Type
+	options                   []string
+	takeover                  bool
+	subresourceGVKMapper      GroupVersionKindProvider
+	validationIdentifier      string
+	normalizationRules        []field.NormalizationRule
+	containsDeclarativeNative bool
 }
 
 // validateDeclaratively validates obj and oldObj against declarative
@@ -344,10 +344,10 @@ func metricIdentifier(ctx context.Context, obj runtime.Object, opType operation.
 }
 
 // ValidateDeclarativelyWithMigrationChecks encapsulates the logic for running declarative validation.
-// It proceeds if either the DeclarativeValidation feature gate is enabled or `containsDeclarativeOnly` is set.
+// It proceeds if either the DeclarativeValidation feature gate is enabled or `containsDeclarativeNative` is set.
 // The function generates a validation identifier, runs declarative validation,
 // and conditionally compares results with imperative validation and merges errors based on the feature gate and `takeover` flag.
-// Declarative-only errors are always appended to the result.
+// Declarative-native errors are always appended to the result.
 func ValidateDeclarativelyWithMigrationChecks(ctx context.Context, scheme *runtime.Scheme, obj, oldObj runtime.Object, errs field.ErrorList, opType operation.Type, configOpts ...ValidationConfig) field.ErrorList {
 	declarativeValidationEnabled := utilfeature.DefaultFeatureGate.Enabled(features.DeclarativeValidation)
 	takeover := utilfeature.DefaultFeatureGate.Enabled(features.DeclarativeValidationTakeover)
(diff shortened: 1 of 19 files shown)


staging/src/k8s.io/apiserver/pkg/registry/rest/validate.go
```diff
@@ -75,13 +75,24 @@ func WithNormalizationRules(rules []field.NormalizationRule) ValidationConfig {
 	}
 }
 
+// WithContainsDeclarativelyOnly marks the validation configuration to indicate that it includes
```
aaron-prindle: This comment is no longer accurate. With the new changes, declarative validation can still run if the feature gate is off but config.containsDeclarativeOnly is true. Please update the comment to reflect this exception.

staging/src/k8s.io/apiserver/pkg/registry/rest/validate.go
```diff
@@ -75,13 +75,24 @@ func WithNormalizationRules(rules []field.NormalizationRule) ValidationConfig {
...
+// declarative validations that are defined *only* declaratively, lacking corresponding imperative validation.
+// This ensures that declarative validations are always executed. Declarative-only validation errors
```
aaron-prindle: nit: currently the doc comment here:
```
// This ensures that declarative validations are always executed. Declarative-only validation errors
// are filtered from the full set of declarative validation errors and returned alongside other validation errors.
```

implies that this option performs the filtering, but it sets the flag that enables the execution and filtering logic within ValidateDeclarativelyWithMigrationChecks.

Might make sense to change to be:

```
// When set, declarative validation is always executed regardless of feature gates. Errors marked as
// declarative-only are separated from the full set and returned alongside imperative errors.
```

staging/src/k8s.io/apiserver/pkg/registry/rest/validate.go
```diff
@@ -75,13 +75,24 @@ func WithNormalizationRules(rules []field.NormalizationRule) ValidationConfig {
...
+// are filtered from the full set of declarative validation errors and returned alongside other validation errors.
+func WithContainsDeclarativelyOnly() ValidationConfig {
```
aaron-prindle: nit: WithContainsDeclarativelyOnly name is a bit hard to parse.  I am ok with it but wanted to note there might be a more clear alternative

staging/src/k8s.io/apiserver/pkg/registry/rest/validate.go
lalitc375: done.

staging/src/k8s.io/apiserver/pkg/registry/rest/validate.go
```diff
@@ -358,14 +368,34 @@ func ValidateDeclarativelyWithMigrationChecks(ctx context.Context, scheme *runti
...
+
 	// Call the panic-safe wrapper with the real validation function.
 	declarativeErrs := panicSafeValidateFunc(validateDeclaratively, cfg.takeover, cfg.validationIdentifier)(ctx, scheme, obj, oldObj, cfg)
 
-	compareDeclarativeErrorsAndEmitMismatches(ctx, errs, declarativeErrs, takeover, validationIdentifier, cfg.normalizationRules)
-	if takeover {
-		errs = append(errs.RemoveCoveredByDeclarative(), declarativeErrs...)
+	mirroredDVErrors := field.ErrorList{}
+	dvOnlyErrors := field.ErrorList{}
+	if cfg.containsDeclarativeOnly {
```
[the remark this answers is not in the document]
lalitc375: It is same case as mentioned in the last comment, I have handled it. DV only validation will return error if DV panicked.

staging/src/k8s.io/apiserver/pkg/registry/rest/validate.go
lalitc375: Open for new ideas.  Don't have a better alternative.

staging/src/k8s.io/apiserver/pkg/registry/rest/validate_test.go
```diff
@@ -793,3 +796,123 @@ func TestMetricIdentifier(t *testing.T) {
...
 }
+
+func TestValidateDeclarativelyWithMigrationChecks(t *testing.T) {
+	// Create errors for testing
+	errImperative := field.ErrorList{
+		field.Invalid(field.NewPath("spec", "restartPolicy"), "imp", "imperative error"),
+		field.Forbidden(field.NewPath("spec", "restartPolicy"), "imperative error covered by declarative").MarkCoveredByDeclarative(),
+	}
+	errMirroredDeclarative := field.Invalid(field.NewPath("spec", "restartPolicy"), "dec", "declarative error")
+	errDeclarativeOnly := field.Invalid(field.NewPath("spec", "replicas"), "decOnly", "declarative only error").MarkDeclarativeOnly()
```
aaron-prindle: nit: do we still want to use *Only vs *Native here?

staging/src/k8s.io/apiserver/pkg/registry/rest/validate_test.go
```diff
@@ -793,3 +796,123 @@ func TestMetricIdentifier(t *testing.T) {
...
+
+	testCases := []struct {
+		name                    string
+		dvFeatureEnabled        bool
+		takeoverEnabled         bool
+		containsDeclarativeOnly bool
```
aaron-prindle: nit: do we still want to use *Only vs *Native here?

jpbetz:
/approve

Leaning on @aaron-prindle's review here

PR #226, opened 2025-12-09
Enable DV native for workload API

lalitc375:
This PR show how DV only will be enabled for an API. How its tests will be written.

The first commit changes testing utilities to test DV which has DV native by comparing the DV native marking field. 

The second commit, enables DV only for the workload API, also delete handwritten code for a field and mark it DV only and update the corresponding test.

### DIFF
--- a/pkg/registry/scheduling/workload/strategy_test.go
+++ b/pkg/registry/scheduling/workload/strategy_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package workload
 
 import (
+	"context"
 	"testing"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


pkg/registry/scheduling/workload/declarative_validation_test.go
```diff
@@ -86,15 +86,15 @@ func testDeclarativeValidate(t *testing.T, apiVersion string) {
 				obj.Spec.PodGroups[0].Name = ""
 			}),
 			expectedErrs: field.ErrorList{
-				field.Required(field.NewPath("spec", "podGroups").Index(0).Child("name"), ""),
+				field.Required(field.NewPath("spec", "podGroups").Index(0).Child("name"), "").MarkDeclarativeNative(),
```
jpbetz: What alternatives did we consider for this?  I can see the upside to being explicit about which tests are for DV-only, but it also seems tedious to have to mark each error individually.  Did we consider any alternatives?  Why did we reject them?

lalitc375: The other alternative is to create a  new test or test cases block for the declarative native cases.  The major advantage of the current setup is the ability to mix and match DV-native and non native validations within a test case. This is also providing a unified view of what all is being tested.

pkg/registry/scheduling/workload/strategy_test.go
```diff
@@ -55,6 +55,13 @@ func TestWorkloadStrategy(t *testing.T) {
 func TestPodSchedulingStrategyCreate(t *testing.T) {
 	t.Run("simple", func(t *testing.T) {
 		ctx := genericapirequest.NewDefaultContext()
+		ctx = genericapirequest.WithRequestInfo(ctx, &genericapirequest.RequestInfo{
```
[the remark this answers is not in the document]
lalitc375: Earlier, DV was throwing errors(Internal errors), but the errors were ignored because the takeover flag is off by default. But now after enabling native flags, we no longer ignore DV errors of the internal type or errors originating from native validations.

aaron-prindle:
/lgtm
merging

PR #236, opened 2026-01-15
(feat): add minLength declarative validation tag

### LINKED ISSUE #235: Add +k8s:minLength tag
Issue tracks adding a +k8s:minLength tag. This tag is used to enforce a minimum length on the size of a string. It should function similarly to other limit tags like +k8s:maxLength

  Expected implementation locations:

  validation method logic:
   * staging/src/k8s.io/apimachinery/pkg/api/validate/limits.go (Add MinLength function)
   * staging/src/k8s.io/apimachinery/pkg/api/validate/limits_test.go

  validators logic:
   * staging/src/k8s.io/code-generator/cmd/validation-gen/validators/limits.go (Update to parse the new tag)

  validation tag test:
   * staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/tags/minlength (New directory for
     generation tests)

  minLength is semantically identical to the minLength Open API validation.

  Why is this needed?

  This tag is needed as minLength is a part of the Open API validation that Declarative Validation needs to
  support, and k/k types use this validation to bound the size of a string.

Related comment here from a k/k PR where such a tag would be useful: https://github.com/kubernetes/kubernetes/pull/135196#discussion_r2695303781

everettraven:
#### What type of PR is this?

/kind feature

Adds a declarative validation tag, `k8s:minLength`, for specifying the minimum allowed length of a string value.

fixes #235


staging/src/k8s.io/apimachinery/pkg/util/validation/field/errors.go
```diff
@@ -173,6 +173,10 @@ const (
 	ErrorTypeInternal ErrorType = "InternalError"
 	// ErrorTypeTypeInvalid is for the value did not match the schema type for that field
 	ErrorTypeTypeInvalid ErrorType = "FieldValueTypeInvalid"
+	// ErrorTypeTooShort is used to report that the given value is too short.
+	// This is similar to ErrorTypeInvalid, but the error will not include the
+	// too-short value. See TooShort().
+	ErrorTypeTooShort ErrorType = "FieldValueTooShort"
```
everettraven: Note: I added this to match the `MaxLength` validation usage of `ErrorTypeTooLong`. We could also use `ErrorTypeInvalid`.

I'm not strongly opinionated here, but adding `ErrorTypeTooShort` seemed like a reasonable path to me.

staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/native/basics/doc.go
```diff
@@ -55,7 +55,6 @@ type MyObject struct {
 
 	// +k8s:declarativeValidationNative
 	// +k8s:maxLength=60
-	// +k8s:minLength=50
```
everettraven: Removing the `+k8s:minLength` marker usage here because it never existed and the `+k8s:declarativeValidationNative` results in generation failures when introducing `k8s:minLength` as `Alpha`.

staging/src/k8s.io/apimachinery/pkg/util/validation/field/errors.go
```diff
@@ -362,6 +368,29 @@ func InternalError(field *Path, err error) *Error {
...
+// error will not include the too-shoort value. If minLength is non-negative, it will
+// be included in the message. The value argument is not used.
+func TooShort(field *Path, _ interface{}, minLength int) *Error {
+	var msg string
+	if minLength >= 0 {
+		bs := "bytes"
+		if minLength == 1 {
+			bs = "byte"
+		}
+		msg = fmt.Sprintf("must be more than %d %s", minLength, bs)
```
aaron-prindle: IIUC this should read "must be at least" (vs. "must be more than")  as it is >= vs >.

PR #237, opened 2026-01-21
Implement declarative validation discriminator and member tags

aaron-prindle:
PR implements declarative modal validation (+k8s:discriminator and +k8s:member)

This feature introduces state-based validation to declarative validation, allowing different validation rules to be applied to a field based on the value of a sibling "discriminator" field.

### Key changes:

- Added `ExtractTagValidations` to the validator registry. This allows meta-validators to extract payload tags while explicitly ignoring recursive type/field validation passes, preventing duplicate generation and infinite loops.
- Introduced `MultiWrapperFunction` to allow the generator to bundle multiple independent validation constraints (e.g., `+k8s:required` and `+k8s:minLength`) into a single conditional execution block.

#### Tag Implementation
- Implemented `+k8s:discriminator` to identify the controlling field. Supports `string`, `bool`, and `integer` types.
- Implemented `+k8s:member` to map discriminator values to specific validation rules. (pointer-handling logic matches similar existing wrapper tags like `+k8s:subfield`.)
- Added a generator-time check to explicitly deny "two-phase" accumulator tags (like `+k8s:listType`) as payloads, preventing global state mutation from escaping the modal block.

### Example usages (from `doc.go`):

```go
type StrictUnion struct {
	TypeMeta int

	// +k8s:discriminator
	D1 string `json:"d1"`

	// +k8s:member("A")=+k8s:required
	FieldA *string `json:"fieldA,omitempty"`

	// +k8s:member("B")=+k8s:required
	FieldB *string `json:"fieldB,omitempty"`
}
```

```go
type SharedField struct {
	TypeMeta int

	// +k8s:discriminator
	D1 string `json:"d1"`

	// Valid in A and B, implicitly forbidden in C.
	// +k8s:member("A")=+k8s:optional
	// +k8s:member("B")=+k8s:optional
	FieldA *string `json:"fieldA,omitempty"`
}
```

```go
type ChainedValidation struct {
	TypeMeta int

	// +k8s:discriminator
	D1 string `json:"d1"`

	// In mode A, it is required AND must have maxLength 5.
	// +k8s:member("A")=+k8s:required
	// +k8s:member("A")=+k8s:maxLength=5
	FieldA *string `json:"fieldA,omitempty"`
}
```

```go
type ImplicitForbidden struct {
	TypeMeta int

	// +k8s:discriminator
	D1 string `json:"d1"`

	// Field is only mentioned for mode A. Mode B should implicitly forbid it.
	// +k8s:member("A")=+k8s:optional
	FieldA *string `json:"fieldA,omitempty"`
}
```

### DIFF
--- a/staging/src/k8s.io/apimachinery/pkg/api/validate/mode.go
+++ b/staging/src/k8s.io/apimachinery/pkg/api/validate/mode.go
@@ -36,15 +36,16 @@ type ModalRule[T any] struct {
 //
 // It performs ratcheting: if the operation is an Update, and neither the discriminator
 // nor the value (checked via equiv) have changed, validation is skipped.
-func Modal[T any, D comparable, P any](ctx context.Context, op operation.Operation, _ *field.Path, obj, oldObj *P, fldPath *field.Path,
-	getValue func(*P) T, getDiscriminator func(*P) D, equiv MatchFunc[T], defaultValidation ValidateFunc[T], rules []ModalRule[T],
+func Modal[T any, D comparable, P any](ctx context.Context, op operation.Operation, structPath *field.Path,
+	obj, oldObj *P, fieldName string, getMemberValue func(*P) T, getDiscriminator func(*P) D,
+	equiv MatchFunc[T], defaultValidation ValidateFunc[T], rules []ModalRule[T],
 ) field.ErrorList {
-	value := getValue(obj)
+	value := getMemberValue(obj)
 	var oldValue T
 	var oldDiscriminator D
 
 	if oldObj != nil {
-		oldValue = getValue(oldObj)
+		oldValue = getMemberValue(oldObj)
 		oldDiscriminator = getDiscriminator(oldObj)
 	}
 
@@ -54,6 +55,7 @@ func Modal[T any, D comparable, P any](ctx context.Context, op operation.Operati
 		return nil
 	}
 
+	fldPath := structPath.Child(fieldName)
 	dStr := fmt.Sprintf("%v", discriminator)
 	for _, rule := range rules {
 		if rule.Value == dStr {
@@ -63,8 +65,11 @@ func Modal[T any, D comparable, P any](ctx context.Context, op operation.Operati
 			return rule.Validation(ctx, op, fldPath, value, oldValue)
 		}
 	}
+
 	if defaultValidation != nil {
 		return defaultValidation(ctx, op, fldPath, value, oldValue)
 	}
+
 	return nil
+
 }
--- a/staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/tags/mode/zz_generated.validations.go
+++ b/staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/tags/mode/zz_generated.validations.go
@@ -99,7 +99,7 @@ func RegisterValidations(scheme *testscheme.Scheme) error {
 // Validate_ChainedValidation validates an instance of ChainedValidation according
 // to declarative validation rules in the API schema.
 func Validate_ChainedValidation(ctx context.Context, op operation.Operation, fldPath *field.Path, obj, oldObj *ChainedValidation) (errs field.ErrorList) {
-	errs = append(errs, validate.Modal(ctx, op, fldPath, obj, oldObj, fldPath.Child("fieldA"), func(obj *ChainedValidation) *string { return obj.FieldA }, func(obj *ChainedValidation) string { return obj.D1 }, validate.DirectEqualPtr, func(ctx context.Context, op operation.Operation, fldPath *field.Path, obj, oldObj *string) field.ErrorList {
+	errs = append(errs, validate.Modal(ctx, op, fldPath, obj, oldObj, "fieldA", func(obj *ChainedValidation) *string { return obj.FieldA }, func(obj *ChainedValidation) string { return obj.D1 }, validate.DirectEqualPtr, func(ctx context.Context, op operation.Operation, fldPath *field.Path, obj, oldObj *string) field.ErrorList {
 		errs := field.ErrorList{}
 		errs = append(errs, validate.ForbiddenPointer(ctx, op, fldPath, obj, oldObj)...)
 		return errs
@@ -131,7 +131,7 @@ func Validate_ChainedValidation(ctx context.Context, op operation.Operation, fld
 // Validate_Collections validates an instance of Collections according
 // to declarative validation rules in the API schema.
 func Validate_Collections(ctx context.Context, op operation.Operation, fldPath *field.Path, obj, oldObj *Collections) (errs field.ErrorList) {
-	errs = append(errs, validate.Modal(ctx, op, fldPath, obj, oldObj, fldPath.Child("listField"), func(obj *Collections) []string { return obj.ListField }, func(obj *Collections) string { return obj.D1 }, validate.SemanticDeepEqual, func(ctx context.Context, op operation.Operation, fldPath *field.Path, obj, oldObj []string) field.ErrorList {
+	errs = append(errs, validate.Modal(ctx, op, fldPath, obj, oldObj, "listField", func(obj *Collections) []string { return obj.ListField }, func(obj *Collections) string { return obj.D1 }, validate.SemanticDeepEqual, func(ctx context.Context, op operation.Operation, fldPath *field.Path, obj, oldObj []string) field.ErrorList {
 		errs := field.ErrorList{}
 		errs = append(errs, validate.ForbiddenSlice(ctx, op, fldPath, obj, oldObj)...)
 		return errs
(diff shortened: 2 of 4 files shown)


staging/src/k8s.io/code-generator/cmd/validation-gen/validators/mode.go
```diff
@@ -0,0 +1,482 @@
...
+		if rules.member != context.Member {
+			return Validations{}, fmt.Errorf("internal error: member mismatch for field %q", fieldName)
+		}
+	} else {
+		group.members[fieldName] = &fieldModeRules{
+			member: context.Member,
+		}
+	}
+
+	payloadValidations, err := mtv.validator.ExtractValidations(context, *tag.ValueTag)
```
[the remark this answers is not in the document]
aaron-prindle: Ah good point thanks.  I've added this method to my PR for now and used it instead of ExtractValidations

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/mode.go
```diff
@@ -0,0 +1,482 @@
...
+
+const (
+	discriminatorTagName = "k8s:discriminator"
+	memberTagName        = "k8s:member"
+)
+
+func init() {
+	RegisterTagValidator(&modeTagValidator{modeDefinitions})
+	RegisterTagValidator(&memberTagValidator{modeDefinitions, nil})
+	RegisterTypeValidator(&modeTypeOrFieldValidator{modeDefinitions})
```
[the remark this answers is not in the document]
aaron-prindle: You are correct. While it was partially working as a TypeValidator (since it inspected the struct's fields during the type pass), it should also be a FieldValidator. I have updated the init() function to register modeTypeOrFieldValidator as both a TypeValidator and a FieldValidator now.

staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/tags/mode/zz_generated.validations.go
```diff
@@ -0,0 +1,451 @@
...
+	// field ListTypeInsideMode.Items
+	errs = append(errs,
+		func(fldPath *field.Path, obj, oldObj []ListItem, oldValueCorrelated bool) (errs field.ErrorList) {
+			// don't revalidate unchanged data
+			if oldValueCorrelated && op.Type == operation.Update && equality.Semantic.DeepEqual(obj, oldObj) {
+				return nil
+			}
+			// call field-attached validations
+			// lists with map semantics require unique keys
+			errs = append(errs, validate.Unique(ctx, op, fldPath, obj, oldObj, func(a ListItem, b ListItem) bool { return a.Name == b.Name })...)
```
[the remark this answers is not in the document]
aaron-prindle: +1, I agree this should be in a modal block.  Discussed offline, will reply when a fix is made - need to think through best way to  do this

staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/tags/mode/doc_test.go
```diff
@@ -0,0 +1,158 @@
...
+	st.Value(&SharedField{D1: "B"}).ExpectValid()
+	st.Value(&SharedField{D1: "B", FieldA: ptr.To("val")}).ExpectValid()
+
+	// Forbidden in C
+	st.Value(&SharedField{D1: "C", FieldA: ptr.To("val")}).ExpectMatches(field.ErrorMatcher{}.ByType().ByField(), field.ErrorList{
+		field.Forbidden(field.NewPath("fieldA"), ""),
+	})
+}
+
+func TestChainedValidation(t *testing.T) {
```
lalitc375: Should we add a test for D1:"B"?

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/mode.go
```diff
@@ -0,0 +1,501 @@
...
+	group := mtv.shared[context.ParentPath.String()].getOrCreate(modeName)
+	if group.discriminatorMember != nil && group.discriminatorMember != context.Member {
+		return Validations{}, fmt.Errorf("multiple fields marked %q with the same name %q in the same struct", discriminatorTagName, modeName)
+	}
+	group.discriminatorMember = context.Member
+
+	return Validations{}, nil
+}
+
+func isInteger(t *types.Type) bool {
```
lalitc375: Gengo has this method.   That has also has bytes.

https://pkg.go.dev/k8s.io/gengo/types#IsInteger

aaron-prindle: Ah thanks.  In going through this I realized this logic isn't necessary anymore with some changes I made related to pointers.  This method is no longer needed now.

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/mode.go
```diff
@@ -0,0 +1,447 @@
...
+	for name := range groups {
+		groupNames = append(groupNames, name)
+	}
+	slices.Sort(groupNames)
+
+	for _, gn := range groupNames {
+		group := groups[gn]
+		if group.discriminatorMember == nil {
+			if len(group.members) > 0 {
+				return Validations{}, fmt.Errorf("missing discriminator for group %q", gn)
```
lalitc375: It will endup printing "default" if  the modal doesn't have name.

staging/src/k8s.io/apimachinery/pkg/api/validate/mode.go
```diff
@@ -0,0 +1,50 @@
...
+// ModalRule defines a validation to apply for a specific mode value.
+type ModalRule[T any] struct {
+	Value      string
+	Validation ValidateFunc[T]
+}
+
+// Modal validates a value based on a discriminator value.
+// It iterates through the rules and applies the first one that matches the discriminator.
+// If no rule matches, it applies the defaultValidation if provided.
+func Modal[T any, D any, P any](ctx context.Context, op operation.Operation, _ *field.Path, obj, oldObj P, fldPath *field.Path, value, oldValue T, discriminator D, defaultValidation ValidateFunc[T], rules []ModalRule[T]) field.ErrorList {
```
lalitc375: _ *field.Path, obj, oldObj P

I don't think these arguments are used.

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/mode.go
aaron-prindle: Ah good catch, fixed now.

staging/src/k8s.io/code-generator/cmd/validation-gen/validators/mode.go
```diff
@@ -0,0 +1,447 @@
...
+			Fields: []StructLiteralField{
+				{Name: "Value", Value: val},
+				{Name: "Validation", Value: wrapper},
+			},
+		})
+	}
+
+	modalValidator := types.Name{Package: libValidationPkg, Name: "Modal"}
+
+	var rulesSlice any = SliceLiteral{
```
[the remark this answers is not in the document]
aaron-prindle: It should be SliceLiteral, not any.  Fixed now, thanks.

staging/src/k8s.io/apimachinery/pkg/api/validate/mode.go
aaron-prindle: Fixed now, thanks.  I initially went a different route than subfield which is how I ended up w/ this issue BUT now they are actually the same w/ a more standard signature (those were there to match the signature needed).

aaron-prindle:
> I don't see ratcheting
Doh thanks, added ratcheting support in the latest commit and associated testing (`TestRatcheting`)

```go
	if op.Type == operation.Update && oldObj != nil && discriminator == oldDiscriminator && equiv(value, oldValue) {
		return nil
	}

```

staging/src/k8s.io/apimachinery/pkg/api/validate/mode.go
```diff
@@ -0,0 +1,70 @@
...
+	Validation ValidateFunc[T]
+}
+
+// Modal validates a value based on a discriminator value.
+// It iterates through the rules and applies the first one that matches the discriminator.
+// If no rule matches, it applies the defaultValidation if provided.
+//
+// It performs ratcheting: if the operation is an Update, and neither the discriminator
+// nor the value (checked via equiv) have changed, validation is skipped.
+func Modal[T any, D comparable, P any](ctx context.Context, op operation.Operation, _ *field.Path, obj, oldObj *P, fldPath *field.Path,
```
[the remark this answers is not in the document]
aaron-prindle: I refactored this once more, now the arg is used and the code is cleaner (prev was deriving this value from a diff field but can actually just use it here)

aaron-prindle:
> Feel free to merge this PR, after addressing the comments.

Merging 👍

PR #245, opened 2026-02-04
Fix +k8s:subfield and conditional wrapper (+k8s:ifEnabled, etc.) leakage by introducing the Context Sandbox Architecture

### LINKED ISSUE #244: 2-Phase Stateful Validation Tags Leak Out of Conditional Wrappers (ex: +k8s:ifEnabled=+k8s:upgrade=NoModify)
When defining validation logic using the `validation-gen` code generator, certain complex "2-phase" stateful tags (like `+k8s:listType=map`, `+k8s:listMapKey`, and `+k8s:update`) do not behave correctly when nested inside conditional wrapper tags (like `+k8s:ifEnabled`, `+k8s:each*`, or the forthcoming `+k8s:member`).

Instead of isolating their logic inside the wrapper's conditional block, the stateful tags "leak" their configuration into global memory. This causes the generated validation rules to be applied universally to the entire field, completely bypassing the user's intended conditional logic.

## Steps to Reproduce
**Branch w/ repro:** https://github.com/aaron-prindle/validation-gen/tree/issue-244-repro

Attempt to apply conditional immutability to a field using the `+k8s:ifEnabled` feature gate wrapper:

```go
type IfEnabledUpdateStruct struct {
	TypeMeta int     `json:"typeMeta"`
	// +k8s:ifEnabled(StrictUpdates)=+k8s:update=NoModify
	Status   *string `json:"status"`
}
```

### Expected Behavior
The generated validation code should contain a block that safely applies the update constraint *only* inside the `validate.IfOption` wrapper:
```go
			// call field-attached validations
			earlyReturn := false
			if e := validate.IfOption(ctx, op, fldPath, obj, oldObj, "StrictUpdates", true, func(ctx context.Context, op operation.Operation, fldPath *field.Path, obj, oldObj *string) field.ErrorList {
				return validate.UpdatePointer(ctx, op, fldPath, obj, oldObj, validate.NoModify)
			}); len(e) != 0 {
				errs = append(errs, e...)
				earlyReturn = true
			}
			if earlyReturn {
				return // do not proceed
			}
```

### Actual Behavior
Because `+k8s:update` writes its state globally and bypasses the wrapper's execution context, the code generator emits a "double validation" leak:

1. The wrapper generates correctly.
2. The global `updateFieldValidator` reads the leaked state and generates a **second**, unconditional `validate.UpdatePointer` check immediately afterwards!

**The generated code looks like this:**
```go
			// 1. Correctly wrapped output (from the wrapper execution)
			if e := validate.IfOption(ctx, op, fldPath, obj, oldObj, "StrictUpdates", true, func(ctx context.Context, op operation.Operation, fldPath *field.Path, obj, oldObj *string) field.ErrorList {
				return validate.UpdatePointer(ctx, op, fldPath, obj, oldObj, validate.NoModify)
			}); len(e) != 0 {
				errs = append(errs, e...)
				earlyReturn = true
			}
            
			// 2. INCORRECT global output (The Leak)
			if e := validate.UpdatePointer(ctx, op, fldPath, obj, oldObj, validate.NoModify); len(e) != 0 {
				errs = append(errs, e...)
				earlyReturn = true
			}
```

aaron-prindle:
fixes #244

This PR fixes a code-generation bug where "2-phase" stateful tags (like `+k8s:listType=map` and `+k8s:update`) would "leak" out of conditional wrapper tags (like `+k8s:subfield`, `+k8s:each*`, and `+k8s:ifEnabled`), applying their validation rules globally rather than conditionally. Fixing this is important to properly support the pattern of tags that accept state-accumulating payload tags.

To solve this, the PR extends validation-gen to support a **Context Sandbox Architecture**, completely unifying how plugins manage state across both local and global execution scopes.

### The Bug (#244)
Previously, stateful tags mutated a global registry (e.g., `globalListMeta` in `list.go`). If an API author tried to restrict update constraints to a specific subfield like so:
```go
// +k8s:subfield(status)=+k8s:update=NoModify
// +k8s:subfield(status)=+k8s:update=NoSet
Config struct { ... }
```
The inner tags (1st phase) would record their state globally and return no code to the `subfield` wrapper, resulting in an empty conditional block. Later, the global FieldValidator (2nd phase) would emit the composite `validate.UpdatePointer` check universally at the root struct level, ignoring the `subfield` targeting completely.

### The Fix: Context Sandbox Architecture
We have upgraded the core generator engine to support isolated state accumulation and eliminated plugin-specific global maps entirely:

1. **Encapsulated Sandbox:** The `Context` struct now includes a strongly-typed, generic `Sandbox` (`validators.GetFromSandbox[T]`, `validators.SetInSandbox[T]`) for plugins to accumulate state safely and avoid collisions by using unique unexported struct keys (similar to `context.Context`).
2. **Unified State Execution:** The `globalRegistry` now owns all root sandboxes and automatically injects them into the `Context`. Plugins blindly read/write to the Sandbox using the generic helpers and no longer care if they are executing globally or inside a wrapper. Legacy plugin maps (`globalListMeta`, `byFieldPath`) have been deleted.
3. **Grouped Execution:** Wrapper tags (like `+k8s:subfield`) now implement `GroupedTagValidator`. The generator groups all payloads for a specific wrapper instance and passes them in a single batch.
4. **Local Phase 2:** The wrapper tag initializes a *new* isolated Sandbox, executes all its payloads within it, and then instantly runs the Phase 2 `FieldValidators` against that local sandbox. 

This ensures that the generated validation logic (even complex, aggregated logic like multiple update constraints) is strictly contained and returned to the wrapper tag to be placed inside the correct conditional block.

### Why this is necessary for `+k8s:member` (Modes)
In a subsequent PR, we'll add DV "Mode" Support (`+k8s:discriminator` and `+k8s:member`), which allow API authors to define validations that only run when a resource is in a specific state. 

This enables conditional immutability as a native Kubernetes pattern (e.g., "Field X is mutable in Mode A, but locked in Mode B"). Without the Sandbox Architecture, an API author writing `// +k8s:member("B")=+k8s:update=NoModify` would permanently lock their field across *all* modes due to the global state leak. 

## Changes
* Added `Sandbox` struct and generic `GetFromSandbox`/`SetInSandbox` helpers to `validators.go`.
* Upgraded `subfield`, `each`, `item`, and `options` wrappers to initialize and execute within a Sandbox.
* Upgraded `listType`, `listMapKey`, `unique`, and `update` tags to use the Sandbox exclusively.
* Deleted plugin-specific global maps, migrating all state to the unified registry sandboxes.
* Fixed non-deterministic output generation for grouped `subfield` evaluations.
* Added test cases to `output_tests/tags/subfield/shallow` proving that composite list-map keys and multiple update constraints can be safely aggregated and executed within nested subfields.

## Related Issues
Prerequisite for `+k8s:discriminator` and `+k8s:member` tags.

### DIFF
--- a/staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/tags/subfield/shallow/doc.go
+++ b/staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/tags/subfield/shallow/doc.go
@@ -56,3 +56,54 @@ type OtherStruct struct {
 type SmallStruct struct {
 	StringField string `json:"stringField"`
 }
+
+type ListItem struct {
+	Name string `json:"name"`
+	Val  string `json:"val"`
+}
+
+// +k8s:subfield(listTypeMap)=+k8s:listType=map
+// +k8s:subfield(listTypeMap)=+k8s:listMapKey=name
+// +k8s:subfield(listTypeSet)=+k8s:listType=set
+type ListStruct struct {
+	ListTypeMap []ListItem `json:"listTypeMap"`
+	ListTypeSet []string   `json:"listTypeSet"`
+}
+
+type ListInsideSubfield struct {
+	TypeMeta int `json:"typeMeta"`
+
+	Lists ListStruct `json:"lists"`
+}
+
+// +k8s:subfield(stringField)=+k8s:update=NoModify
+type UpdateStruct struct {
+	StringField string `json:"stringField"`
+}
+
+type UpdateInsideSubfield struct {
+	TypeMeta int `json:"typeMeta"`
+
+	Updatable UpdateStruct `json:"updatable"`
+}
+
+// AGGREGATION TEST:
+// The Sandbox architecture allows grouping multiple 'listMapKey' tags inside the same wrapper.
+
+// +k8s:subfield(listTypeMap)=+k8s:listType=map
+// +k8s:subfield(listTypeMap)=+k8s:listMapKey=name
+// +k8s:subfield(listTypeMap)=+k8s:listMapKey=val
+type DuplicateAccumulatorStruct struct {
+	TypeMeta    int        `json:"typeMeta"`
+	ListTypeMap []ListItem `json:"listTypeMap"`
+}
+
+// AGGREGATION TEST:
+// The Sandbox architecture allows grouping multiple 'update' constraints inside the same wrapper.
+
+// +k8s:subfield(stringField)=+k8s:update=NoModify
+// +k8s:subfield(stringField)=+k8s:update=NoUnset
+type AggregatedUpdateStruct struct {
+	TypeMeta    int     `json:"typeMeta"`
+	StringField *string `json:"stringField"`
+}
--- a/staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/tags/subfield/deep/zz_generated.validations.go
+++ b/staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/tags/subfield/deep/zz_generated.validations.go
@@ -62,15 +62,12 @@ func Validate_Struct(ctx context.Context, op operation.Operation, fldPath *field
 				return nil
 			}
 			// call field-attached validations
-			func() { // cohort structField
-				errs = append(errs, validate.Subfield(ctx, op, fldPath, obj, oldObj, "structField", func(o *OtherStruct) *SmallStruct { return &o.StructField }, validate.DirectEqualPtr, func(ctx context.Context, op operation.Operation, fldPath *field.Path, obj, oldObj *SmallStruct) field.ErrorList {
-					return validate.Subfield(ctx, op, fldPath, obj, oldObj, "stringField", func(o *SmallStruct) *string { return &o.StringField }, validate.DirectEqualPtr, func(ctx context.Context, op operation.Operation, fldPath *field.Path, obj, oldObj *string) field.ErrorList {
-						return validate.FixedResult(ctx, op, fldPath, obj, oldObj, false, "Struct.StructField.StructField 1")
-					})
-				})...)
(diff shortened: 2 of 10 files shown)


lalitc375 (commented):
It is a good direction.

1. I think we can eliminate global states and switch everything to use the sandbox. Tags without +k8s:subfield can also rely on the sandboxes. Then we would not need two models(Global states and sandboxes).  Let me know what you think about this idea.


2. The key in the sandbox is not playing a major role, it is constant for a validator. If we are creating a sandbox for a validator, then sandbox can be simplified further.

aaron-prindle:
Refactored the commits to be more readable given new global sandbox architecture (vs N refactors)

staging/src/k8s.io/code-generator/cmd/validation-gen/output_tests/tags/subfield/shallow/doc.go
```diff
@@ -56,3 +56,54 @@ type OtherStruct struct {
 type SmallStruct struct {
 	StringField string `json:"stringField"`
 }
+
+type ListItem struct {
+	Name string `json:"name"`
+	Val  string `json:"val"`
+}
+
+// +k8s:subfield(listTypeMap)=+k8s:listType=map
```
thockin: I can see how this breaks.  Once you apply https://github.com/kubernetes/kubernetes/pull/136284#discussion_r2766388112 then it doesn't hard-fail but it doesn't emit code either.  That's because the `listValidator` is attached to  fields and types, but it doesn't descent INTO the fields OF the type.

Maybe what we need is something like a task-queue, where phase-1-of-2 tags can put an indicator that there's work to be done (they have the context, so they know if it is a field or type, the path, and they know which validator has to run).

This could even eliminate the need for FieldValidators and TypeValidators?  I am not sure -- unions still need it.  It could eliminate the need for some (e.g. the `listValidator`) Field/TypeValidators.

aaron-prindle:
Thanks for the reviews here, going to close this for now as Lalit's PR upstream solves some of the issues identified.  We still have issues with aggregated tags so we might want to revisit this solution but it can wait post v1.36.  I will update my issue to explain the current status

